PR #1, opened 2017-10-12
Updating and cleaning npm packages, updating the webpack build, updat…

EdwinGuzman:
…ing the readme, adding unit tests, coverage, and an example.


.nyc_output/28b6eec849100fb8405b1db8caf0bed3.json
```diff
@@ -0,0 +1 @@
+{}
```
rhernand3z: Seems like this wasn't gitignored..

package.json
```diff
@@ -81,11 +85,19 @@
     "react-tools": "0.13.3",
     "react-addons-css-transition-group": "0.14.8",
```
rhernand3z: Wouldn't this dependency conflict with the current react v15.x?

EdwinGuzman: Doesn't conflict and it's a dependency I hope we can remove soon.

.nyc_output/28b6eec849100fb8405b1db8caf0bed3.json
EdwinGuzman: Updated.

PR #4, opened 2017-10-13
Removing unneeded components, small syntax updates here and there, re…

EdwinGuzman:
…placing the older css transition npm package for a newer one.

More random refactoring but mostly removing old and unneeded components, also conforming to more eslint Airbnb style syntax.


src/app/utils/PickItemModel.js
```diff
@@ -16,22 +21,22 @@ function PickItemModel() {
...
 
     if (!data) {
       return item;
     }
 
     item.id = data.id;
     item.type = data.type;
-    
+
     item.title = data.attributes.title;
```
rhernand3z: I would wrap all these in a if (data.attributes) conditional check because you will get a `cannot read catalog-slug property of undefined` if `data.attributes doesn't exist. I don't see that check...

EdwinGuzman: These files will go once we get a new data model so I rather not spend time updating these.

PR #5, opened 2017-10-16
Breaking up server Express routes functions.

EdwinGuzman:
This is an attempt to break out all the functions in the old `ApiRoutes` into cleaner and separate "month/season" and "annual" data fetching functions. Those functions are associated to Express routes and I tried to make that more obvious.

### DIFF
--- a/src/server/routes/annualData.js
+++ b/src/server/routes/annualData.js
@@ -11,6 +11,7 @@ const {
   pageSize,
   includes,
   api,
+  baseUrl,
 } = config;
 
 const appEnvironment = process.env.APP_ENV || 'production';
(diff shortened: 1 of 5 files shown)


src/server/routes/annualData.js
```diff
@@ -0,0 +1,104 @@
...
+
+  if (idOrType === 'childrens') {
+    return annualCurrentData('c100', req, res, next);
+  }
+
+  if (idOrType === 'ya') {
+    return annualCurrentData('ya100', req, res, next);
+  }
+
+  return res.redirect('/books-music-dvds/recommendations/staff-picks/');
```
rhernand3z: Would be ideal to have this url string as part of the config so if there needs to be a change, it can be configurable versus hard-coded.

EdwinGuzman: Good point.

EdwinGuzman:
@rhernand3z updated to baseUrl in the config. Thanks.

PR #6, opened 2017-10-16
Component code cleanup

EdwinGuzman:
Removed `moment, updated props to make it easier to differentiate between month/season and annual pick lists, and also updated components with a config baseUrl to easily update the routes.

@kfriedman you can base your PR off of this change in the react-router routes.


EdwinGuzman:
Yes, agree. The issue is that the new design doesn't have the Age Tabs for the 2017 annual list, but it will be needed for the previous regular month/season lists.

Some things that can be removed would be the grid/list display values from the store and the button that toggle them, as well as the making the modal a separate page. I'm slowly working towards removing it and expect another PR soon.

EdwinGuzman:
I'll try to better outline what should be removed/updated.

PR #7, opened 2017-10-17
App state refactor

EdwinGuzman:
Removed a lot of grid/list code that won't be needed and a component. Added the Store at the top level and the idea will be to bring that data down as props to the components instead of having each component call the store and have its own state. This is just the first part to not make too big PRs.


src/app/components/Application/Application.jsx
```diff
@@ -4,61 +4,88 @@ import PropTypes from 'prop-types';
...
+
+  onChange() {}
+
+  render() {
+    const annualList = !!(this.props.params && this.props.params.type &&
+      (this.props.params.type === 'childrens' || this.props.params.type === 'ya'));
+    let mobileAboutLink = null;
 
-          {mobileAboutLink}
+    if (annualList) {
```
rhernand3z: Would be cleaner to have a function that renders the about mobile link. Would be re-usable if needed and can verify the existence of annual list prop to return. The href should also be part of the config file, in the event it is changed, we have one place to make modifications.

src/app/components/Hero/Hero.jsx
```diff
@@ -90,16 +87,16 @@ class Hero extends React.Component {
...
+        <div className="Hero__container">
+          <div className="Hero__text">
+            <div className="Hero__text__HeroTitle">
               <h3>{heroData.title}</h3>
-              <p className={`${this.props.className}__text__HeroTitle__des`}>
+              <p className="Hero__text__HeroTitle__des">
                 {heroData.description}
               </p>
-              <p className={`${this.props.className}__text__HeroTitle__intro`}>
+              <p className="Hero__text__HeroTitle__intro">
```
rhernand3z: I would propose we pay a re-visit to this later on and establish a class naming convention similar to the toolkit.

EdwinGuzman: Definitely agree.

PR #10, opened 2017-10-18
Updating the react-router version and updating how the main routing base works.

EdwinGuzman:
Version bump and adding a `Main` component so that component routing can work better, and to be easier to remove the `Modal` in the next PR.


server.js
```diff
@@ -61,8 +61,8 @@ app.use('/', (req, res) => {
   alt.bootstrap(JSON.stringify(res.locals.data || {}));
   const iso = new Iso();
 
-  const blogAppUrl = (req.url).indexOf('books-music-dvds/recommendations/staff-picks') !== -1;
-  const routes = blogAppUrl ? appRoutes.client : appRoutes.server;
+  const serverUrl = (req.url).indexOf('books-music-dvds/recommendations/staff-picks') !== -1;
```
rhernand3z: would be better to use the config.baseUrl value from a previous PR instead? less places to track if the url changes.

kfriedman: FYI, I'm working on updating all the routes to use config.

EdwinGuzman: Yup, I purposely left some for @kfriedman

EdwinGuzman:
Fixed conflicts based on @kfriedman 's PR.

PR #11, opened 2017-10-18
Modal removal

EdwinGuzman:
Removed the modal and other unneeded npm packages. Also removed related modal scss.

Updated the modal component into a very simple page view for a book. Example:
screen shot 2017-10-17 at 11 03 27 pm

@ricardoom we'll need a design for this page but it's low priority and a simple view should suffice.


src/app/components/BookContent/BookShare.jsx
```diff
@@ -25,13 +29,13 @@ class BookShare extends React.Component {
 
     nyplStaffPick = `${shareText} ${bookTitle}`;
 
-    this.state = {
+    this.setState({
```
rhernand3z: linter doesn't complain about using setState(); in componentDidMount? https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-mount-set-state.md

Could possibly resolve using their recommended approach if it's a linter rule.

EdwinGuzman: True. This is using `window` for the page's url. So in the next iteration we can remove how that works and then remove `setState` from `componentDidMount`.

EdwinGuzman:
Thanks.

PR #15, opened 2017-10-18
Update routing

kfriedman:
This PR attempts to improve routing by:

- Creating the new `/books-music-dvds/recommendations/best-books/...` route
- Moving hard-coded routes to config
- Separating monthly/annual logic a bit to make things clearer
- Moving the API route to a master base route
- Making naming a bit more consistent

It is *not* a comprehensive overhaul and intended to be more of iterative improvement. Also, I didn't do too much work on the annual logic because I think that needs a lot more thought/refactoring.

There might also be some conflicts now with the `react-router-update` branch.

Anyway, I'm pretty new to a lot of this so please make suggestons/ask questions!

### DIFF
--- a/server.js
+++ b/server.js
@@ -61,8 +61,9 @@ app.use('/', (req, res) => {
   alt.bootstrap(JSON.stringify(res.locals.data || {}));
   const iso = new Iso();
 
-  const blogAppUrl = (req.url).indexOf('books-music-dvds/recommendations/staff-picks') !== -1;
-  const routes = blogAppUrl ? appRoutes.client : appRoutes.server;
+  const isApiRoute = (req.url).indexOf(appConfig.baseApiUrl) !== -1;
+
+  const routes = isApiRoute ? appRoutes.server : appRoutes.client;
 
   match({ routes, location: req.url }, (error, redirectLocation, renderProps) => {
     if (error) {
--- a/src/server/routes/routes.js
+++ b/src/server/routes/routes.js
@@ -2,36 +2,18 @@ import express from 'express';
 
 import monthData from './monthData.js';
 import annualData from './annualData.js';
-import config from '../../../appConfig.js';
+import appConfig from '../../../appConfig.js';
 
 const router = express.Router();
-const { baseUrl } = config;
+const { baseUrl } = appConfig;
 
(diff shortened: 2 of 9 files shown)


server.js
```diff
@@ -61,8 +61,9 @@ app.use('/', (req, res) => {
   alt.bootstrap(JSON.stringify(res.locals.data || {}));
   const iso = new Iso();
 
-  const blogAppUrl = (req.url).indexOf('books-music-dvds/recommendations/staff-picks') !== -1;
-  const routes = blogAppUrl ? appRoutes.client : appRoutes.server;
+  const isApiRoute = (req.url).indexOf(appConfig.baseApiUrl) !== -1;
```
EdwinGuzman: This should be `baseMonthUrl`. This logic will have to be updated now since there are two separate routes.

src/server/routes/routes.js
```diff
@@ -2,64 +2,37 @@ import express from 'express';
...
-  // There are more matches if anything is added after the url slug
-  if (id && id.match(/[0-9]*[-a-z]+/g).length > 1) {
-    return res.redirect(baseUrl);
-  }
-
   return monthData.selectMonthData(req, res, next);
 }
 
 router
-  .route('/')
```
EdwinGuzman: I think we need these routes for the reverse proxy case.

EdwinGuzman: We can add these back in later.

server.js
EdwinGuzman: We should test this once we start doing the reverse proxy. Right now I think I just see "Not Found" pages.

EdwinGuzman:
Worried a bit about the reverse proxy scenario but we can worry about that later. Looks good. I think you can merge this in first and I'll update my PR and fix conflicts.

src/server/routes/routes.js
rhernand3z: Yes I agree, I think we should rethink the whole reverse proxy approach now that we have more AWS control over this and the gateway routing.

EdwinGuzman:
Yes, Kevin's PR actually clears up a lot of questions we have. Having `/staff-picks` and `/best-books` makes it soooo much easier now.

PR #16, opened 2017-10-20
Repo redo

EdwinGuzman:
Sorry for the huge PR but it's hard to break it up.

* Removed a few components, scss files, images and svg icons.
* Updated components to be the bare minimum when it comes to what it renders. So a lot of the logic has been removed.
* Router has been updated and is simpler. Now contains a 404 page too.
* Removed more npm packages that are not needed.

What I didn't do:
* Updated CSS class names.
* Did not touch the existing "modal" component. That's not a priority and there's no way to get to it now. But, this is important to support backwards compatibility for existing urls that are out in the wild.
* Clean up the SCSS files.
* Update the server API endpoint logic. We do not have the new endpoints so I'm keeping the JSON API functions and parser in place until later. I'm thinking of making another PR with the sample JSON that @nonword gave us and a simple endpoint IN the app itself. I can do that next.


rhernand3z:
I was under the assumption that the first StaffPicks app was going to continue to exist with the existing URL that supported the modal. This app essentially was a new route/reverse-proxy to only support the list due Nov 1 with the programming in mind to support other lists. The next iteration of this App would be to kill the existing StaffPicks app running and import all the data into this new App? Not sure why we need a modal component then...

EdwinGuzman:
It's not a modal component. I did "modal" because it's the same route but just a simple page as I showed in a previous PR. Yes, we won't have that for this first iteration but it's easy to implement so I left it there. 

When we kill the old staff picks and import data to this new one, we'll have to support something like `https://www.nypl.org/books-music-dvds/recommendations/staff-picks/annual/childrens/2015-11-01/9780544102293-drum-dream-girl-how-one-girls-courage-changed-music` which should go to something new like `https://www.nypl.org/books-music-dvds/recommendations/best-books/childrens/2015-11-01/9780544102293-drum-dream-girl-how-one-girls-courage-changed-music` (redirects which I don't think we'll have to do too much for).

I'll leave the route but I'll remove the component which just renders the simple individual book data to ease confusion.

PR #19, opened 2017-10-24
First step in dding functionality to be able to filter on the list it…

EdwinGuzman:
…em cards.

This is the first step in filtering and didn't want to do more so this PR wouldn't get too big. It's currently an "or" filtering which was mostly accidental. The next iteration will be an "and" filtering. Removing the filters as you select them is a whole different, huge task, so I'll work on that after this first logic. 

@ricardoom :
screen shot 2017-10-24 at 9 14 15 am

### DIFF
--- a/src/app/components/BookFilters/Filter.jsx
+++ b/src/app/components/BookFilters/Filter.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  CheckSoloIcon,
+  DotsIcon,
+} from 'dgx-svg-icons';
+
+class Filter extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      selected: false,
+      buttonActive: false,
+    };
+    this.onClick = this.onClick.bind(this);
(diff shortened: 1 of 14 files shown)


ricardoom:
@EdwinGuzman good work... I'll ping you & @rhernand3z on whats next..

src/app/components/BookFilters/Filter.jsx
```diff
@@ -0,0 +1,60 @@
...
+    }, 1000);
+  }
+
+  render() {
+    const {
+      selected,
+      buttonActive,
+    } = this.state;
+    const filter = this.props.filter;
+    const active = selected ? 'active' : '';
```
rhernand3z: I would give it `activeClass` name to be clearer it's a class.

EdwinGuzman:
Enzyme, chai, mocha. I want to write tests, but deadline 💀

src/app/components/BookFilters/Filter.jsx
EdwinGuzman: 👌

PR #28, opened 2017-10-24
Update booklist component to use reusable testable functions

rhernand3z:
I noticed that the BookList component was assigning constants and relying on prop checking inside the filtering logic. This quick PR isolates the logic to use functions and filter on function params. In the event we have time to test, it will be a lot easier to test with functions and using params. I Also think that the renderBookItems function might return another component or DOM that displays to the user that the selected filter returns no matches?


EdwinGuzman:
Ideally, we should not get to the point where there are no matches.

PR #30, opened 2017-10-24
Filter 'and' logic

### LINKED ISSUE #25: Update filter logic - "AND" filtering for the item cards

EdwinGuzman:
Fixes #25 

Adds "AND" filtering. Example below based on not-so-good available data:

screen shot 2017-10-24 at 12 12 09 pm

You can try clicking on `adventure` `fantasy` `friendship` `funny`, in any order, and you should get that one book as a result "Pugs of the Frozen North...". If you do not get any results from a filter, it's because the data is not right - those tags there are not necessarily the tags we will used but are hardcoded until we get the right dataset.

### DIFF
--- a/src/app/components/BookList/BookList.jsx
+++ b/src/app/components/BookList/BookList.jsx
@@ -1,28 +1,22 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+
 import Book from '../Book/Book.jsx';
-import {
-  contains as _contains,
-  each as _each,
-} from 'underscore';
+import utils from '../../utils/utils';
 
 const Books = (props) => {
   const currentMonthPicks = props.currentMonthPicks;
   const picks = currentMonthPicks.picks ? currentMonthPicks.picks : [];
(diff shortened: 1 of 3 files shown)


src/app/components/BookList/BookList.jsx
```diff
@@ -1,37 +1,35 @@
...
-              inSelectedFilter.push(bookTag);
-            }
-          });
-          console.log(inSelectedFilter);
-          if (inSelectedFilter.length) {
+        return pickItems.filter(book => {
+          // Get the pick's tags in an ID readable array
+          const tagArray = utils.getPickTags(book);
+          // Get the array of selected tags found in the book item
+          const inSelectedFilter = utils.getSelectedTags(tagArray, props.selectedFilters);
```
rhernand3z: should use params selectedFilters instead of props.selectedFilters

EdwinGuzman: Updated. Thanks.

PR #32, opened 2017-10-25
Removing not available filters

### LINKED ISSUE #26: Update filter logic - remove filter if not applicable

EdwinGuzman:
Fixes #26 Big PR:

* Removes re-useable functions and puts them in `utils`.
* Removes Action calls and instead triggers function calls up to the parent `Main` component.
* This is useful to keep track of both the filtered book item array and the filtered filter array.
* Clicking on a filter updates the SVG and also triggers the focus on them AFTER they get rendered since we want to keep track of what was selected/deselected for a11y purposes.

### DIFF
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -3,21 +3,83 @@ import PropTypes from 'prop-types';
 
 import BookList from '../BookList/BookList.jsx';
 import Sidebar from '../Sidebar/Sidebar.jsx';
+import utils from '../../utils/utils';
 
-const App = (props) => (
-  <div className="nypl-row">
-    <Sidebar filters={props.filters} />
+class App extends React.Component {
+  constructor(props) {
+    super(props);
 
-    <BookList
-      currentMonthPicks={props.currentMonthPicks}
-      selectedFilters={props.selectedFilters}
-    />
-  </div>
-);
+    this.state = {
+      selectableFilters: this.props.selectableFilters,
+      selectedFilters: this.props.selectedFilters,
+      picks: this.props.currentMonthPicks.picks,
+    };
+
+    this.setSelectedFilter = this.setSelectedFilter.bind(this);
+  }
+
+  getNewPickSet(picks, selectedFilters) {
+    if (!selectedFilters.length) {
+      return picks;
+    }
+
+    return picks.filter(book => {
+      // Get the pick's tags in an ID readable array
+      const tagArray = utils.getPickTags(book);
+      // Get the array of selected tags found in the book item
+      const inSelectedFilter = utils.getSelectedTags(tagArray, selectedFilters);
+
+      if (inSelectedFilter.length &&
+          (inSelectedFilter.length === selectedFilters.length)) {
+        return book;
+      }
+
+      return undefined;
+    });
+  }
+
+  setSelectedFilter(filterId, active) {
+    let selectedFilters = [];
+
+    if (active) {
+      selectedFilters = this.state.selectedFilters.concat(filterId);
+    } else {
+      selectedFilters = this.state.selectedFilters.filter(f => f !== filterId);
+    }
+
+    const picks = this.getNewPickSet(this.props.currentMonthPicks.picks, selectedFilters);
+    const selectableFilters = utils.getSelectableTags(picks);
+
+    this.setState({
+      selectableFilters,
+      picks,
+      selectedFilters,
+    });
+  }
+
+  render() {
+    return (
+      <div className="nypl-row">
+        <Sidebar
+          filters={this.props.filters}
+          selectableFilters={this.state.selectableFilters}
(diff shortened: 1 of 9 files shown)


src/app/components/Application/Main.jsx
```diff
@@ -3,21 +3,83 @@ import PropTypes from 'prop-types';
...
+      const tagArray = utils.getPickTags(book);
+      // Get the array of selected tags found in the book item
+      const inSelectedFilter = utils.getSelectedTags(tagArray, selectedFilters);
+
+      if (inSelectedFilter.length &&
+          (inSelectedFilter.length === selectedFilters.length)) {
+        return book;
+      }
+
+      return undefined;
```
rhernand3z: Is returning undefined the right way? Shouldn't it return an empty array being that a pickSet is an array structure? Thoughts?

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -1,38 +1,84 @@
...
 
 class BookFilters extends React.Component {
   constructor(props) {
     super(props);
 
+    this.state = {
+      // Create an array data structure of filter objects.
+      filters: this.props.filters.map(filter => ({
+        active: false,
+        id: filter.toLowerCase().split(' ').join('-'),
```
rhernand3z: In the event that this.props.filters is empty, the initial state assumes it's mappable and might throw an error if it cannot execute `.toLowerCase()` of undefined? Should we check if it is not empty prior to traversal?

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -1,38 +1,84 @@
...
+
     this.renderItems = this.renderItems.bind(this);
     this.onClick = this.onClick.bind(this);
+    this.getFilterArray = this.getFilterArray.bind(this);
   }
 
   onClick(filterId, active) {
-    Actions.setSelectedFilter(filterId, active);
+    // const filterId = filter.toLowerCase().split(' ').join('-');
+    const f = _findWhere(this.state.filters, { id: filterId });
```
rhernand3z: I would promote better naming convention instead of `f` -- following our engineering principles. See https://github.com/NYPL/engineering-general/blob/master/standards/naming-conventions.md#identifiers

src/app/components/BookFilters/Filter.jsx
```diff
@@ -1,60 +1,82 @@
...
-    });
-
+  componentDidMount() {
+    // Since the list of filters get re-rendered, when the a new list is generated and it's
+    // longer than the previous list, then we need to focus on the selected/deselected filter
+    // once all the filters are mounted.
     setTimeout(() => {
-      this.setState({ buttonActive: false });
-    }, 1000);
+      if (this.props.filter.id === this.props.focusId) {
```
rhernand3z: I likes this clean solution!

src/app/components/BookFilters/Filter.jsx
```diff
@@ -1,60 +1,82 @@
...
-    const activeClass = selected ? 'active' : '';
-    const iconType = buttonActive ? <DotsIcon /> : <CheckSoloIcon />;
+      filter,
+      focusId,
+    } = nextProps;
+    // If the focusId, the filter that was JUST selected/deselected, matches the filter,
+    // then load the animation and remove it shortly after. Also focus on it whether it was
+    // selected or deselected for accessibility.
+    if (filter.id === focusId) {
+      this.setState({
```
rhernand3z: can probably be on one line since it updates a single state object property

src/app/components/Application/Main.jsx
EdwinGuzman: This is inside the `filter` method so it's okay to return undefined since it doesn't pass the test. The method itself will return an empty array.

src/app/components/BookFilters/BookFilters.jsx
EdwinGuzman: The default for `this.props.filters` is an empty array. If it's empty, map will return an empty array and won't go through the callback function.

src/app/components/BookFilters/Filter.jsx
EdwinGuzman: Updated.

src/app/components/BookFilters/BookFilters.jsx
EdwinGuzman: Updated. Thanks.

src/app/components/BookFilters/BookFilters.jsx
rhernand3z: 👍

src/app/components/Application/Main.jsx
rhernand3z: roger 👍

rhernand3z:
Let's merge so I can deploy the latest to AWS...

EdwinGuzman:
Done.

PR #40, opened 2017-10-26
Filter button states

EdwinGuzman:
Cleaning up the different states for the buttons by cleaning up the scss and the logic that handles active filters. Now the `BookFilters` component handles the logic and passes the necessary variables. There's a short 400ms delay, or transition state, for the buttons and `Filter` component itself handles that state.

Removed all the colors from `_colors.scss` and added `$nypl-blue-sky`, @ricardoom .


rhernand3z:
I also think it's a good pattern that we adopted in GALC to avoid multiple submissions racing one another.

EdwinGuzman:
Good recommendation. I'll make an issue. 👍

PR #43, opened 2017-10-27
Alt tests

EdwinGuzman:
Removing `dgx-alt-center` and just using `alt`.
Updating actions and stores and cleaning them up. Updating `currentMonthPicks` to `currentPicks` for the store variable name.
Updating the nyc/coverage tool to write tests for the Actions and Store.


EdwinGuzman:
It's slightly easier and also easier to keep up to date without affecting other apps.

PR #45, opened 2017-10-27
Update dom and style book item component

### LINKED ISSUE #22: Modify DOM & Style Book Cards as per design and toolkit
Using a mobile first, a11y first approach to integrate the book cards.

rhernand3z:
Necessary logic applied to DOM structure and styles to achieve design -- https://github.com/NYPL/nypl-bestbooks-2017

Fixes #22 

Note: I was having various issues with nesting the toolkit grid system which caused way too many overrides and was error prone. I went with a simple solution handled by the app bookItem sass.


src/app/components/Book/Book.jsx
```diff
@@ -1,29 +1,70 @@
...
+      <div className="book-item-image-box">
+        <img alt="" src={fullImgSrc} />
+      </div>
+    );
+  };
+
+  const renderTitle = (title, link) => {
+    const titleClass = 'book-item-title';
+
+    if (title && title !== '') {
```
EdwinGuzman: You should be able to just do `if (title) {` since `''` will return a falsy value.

src/app/components/Book/Book.jsx
```diff
@@ -1,29 +1,70 @@
...
-        <img alt="" src={fullImgSrc} />
-      </div>
-      <div className="book-item-description">
-        <h4>
-          <a href="#">{book.title}</a>
-        </h4>
-        <p>{book.text}</p>
-      </div>
+      {renderBookCoverImage(book.imageUrl)}
+      {renderTitle(book.title, '#')}
```
EdwinGuzman: I thought we wanted to link the title to the `catalogUrl` as well?

src/client/styles/utils/_base.scss
```diff
@@ -1,83 +1,83 @@
...
-  width: 1px;
-}
-
-.container {
-  max-width: $container-width;
-  padding: 0.5rem 1rem;
-  margin: 0 auto;
-  @include box-sizing(border-box);
-}
+// a {
```
EdwinGuzman: Remove the code if it's not needed rather than commenting it out.

src/app/components/Book/Book.jsx
rhernand3z: what if title is not of type string, possibly an object or array? maybe I should simply check typeof? I guess I had issues like this with the Hold stuff and needed to be explicit about empty strings... 

```
if (typeof title === 'string' && title.trim() !== '') // better?
```

src/app/components/Book/Book.jsx
rhernand3z: yea but for now since I don't want to do too much work on modifying the sample JSON, I just passed the hash. Once we merge Paul's latest JSON file, I can fix to pass book.encoreUrl or something...

src/client/styles/utils/_base.scss
rhernand3z: 👍

rhernand3z:
@EdwinGuzman -- Added a helper function that strictly checks type and emptyness of book string props. I think it's clearer now the validation of each prop prior to returning the DOM.

PR #48, opened 2017-10-27
Removed react-doc-meta npm module which was the culprit

### LINKED ISSUE #37: Fix Accessing PropTypes Warning from React
Upon running the application, a warning is thrown by React:

```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
Warning: A Component: React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.
```

This needs to be fixed to deliver a bug-free stable application.

CC: @kfriedman @EdwinGuzman

rhernand3z:
- react-doc-meta is outdated since React 0.13ish and does not incorporate PropTypes module, removed from dependencies.
- used appConfig metaTags array to render server-side.
- Tested in production mode, see screenshot
screen shot 2017-10-27 at 5 02 49 pm

Fixes #37 

CC: @RobKelley

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "node-sass": "3.8.0",
     "prop-types": "15.5.10",
     "react": "15.5.4",
-    "react-doc-meta": "0.2.0",
     "react-dom": "15.5.4",
     "react-hot-loader": "1.2.8",
     "react-router": "3.0.5",
--- a/server.js
+++ b/server.js
@@ -10,7 +10,6 @@ import webpackConfig from './webpack.config.js';
 import appConfig from './appConfig.js';
 
 import React from 'react';
-import DocMeta from 'react-doc-meta';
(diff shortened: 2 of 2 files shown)


EdwinGuzman:
To be clear, the module did write the meta tags dynamically on a React page client side and but still updated the server side. So the logic for the correct meta tags was in the React components. If you leave it in the config, that's fine but you'll have to manage it on Express routes instead of React routes.

rhernand3z:
That's fine, the focus of this PR was to remove an outdated library and centralize meta tags which IMO a config file is ideal. I've seen other solutions such as react-helmet that provide meta tags on both client and server and it is more up to date and currently maintained. In the event we need to render tags dynamically, it's a good option. -- https://www.npmjs.com/package/react-helmet

PR #50, opened 2017-10-30
Integrate is js enabled boolean server to client via store

### LINKED ISSUE #46: Pass the JS boolean to the App via State
In order to hide the filters when JS is turned off, we need a way for the client to turn the boolean on when JS is enabled. This uses the same approach as Discovery.

rhernand3z:
Sets up the boolean flag in the Store and updates on componentDidMount via an action. Components that either 1) hide filters when JS is off 2) hide tags in BookItems when JS is on -- will verify the existence of the boolean prop and re-render accordingly.

Fixes #46 

Other tasks/issues will be created to 1) load the filters once JS is on 2) hide the tags once JS is on for individual BookCard items.

CC: @kfriedman @RobKelley @ricardoom


EdwinGuzman:
I forgot, you need to add tests for this new Action and Store value.

PR #51, opened 2017-10-30
About component

### LINKED ISSUE #18: Create About component

EdwinGuzman:
Fixes #18 

Moved the About component from the sidebar to below the list of books. Also added correct link URLs. Need FB and Twitter Icons (@ricadoom) but wanted to make the PR for placement and a11y purposes. The icons can be easily added later on.


src/app/components/About/About.jsx
```diff
@@ -1,15 +1,20 @@
...
-      <li>Facebook icon</li>
-      <li>Twitter icon</li>
-    </ul>
+  <div className="about nypl-row">
+    <div className="inner">
+      <h3>Additional Information</h3>
+      <p>Many of these titles are available in formats
+        for <a href="https://www.nypl.org/accessibility/print-disabilities">patrons
+        with print disabilities.</a></p>
+      <p><a href="https://www.nypl.org/books-music-dvds/recommendations/about/annual-lists">About
```
rhernand3z: you can probably move the <a> to a new line for easier reading. Can these URL's be in the appConfig, in the event of a change, we know all urls are in one central place?

rhernand3z:
We should have the designs updated for a visual reference, the position seems off to me but that's only my opinion. As far as the implementation, it looks good -- minor comments.

src/app/components/About/About.jsx
EdwinGuzman: Cool, updated.

EdwinGuzman:
@wlla  what do you think? This is on dev now.

wlla:
The Additional Information section is marked an `h3` in the mockup. However, I believe it should be an `h2`, making it the same level as the Filter by Tags section and Best Books list. Otherwise, it's buried at the end of the books list.

EdwinGuzman:
@ricardoom  ☝️

EdwinGuzman:
Updated the header. Merging.

PR #57, opened 2017-10-30
Update hero

kfriedman:
This PR:

- Cleans up the Hero component so no data is hard-coded and unnecessary code is removed
- Changes the Hero image to be read from data (eventually the API response)
- Adds a `listMetaData` object to `appConfig`
  - After speaking with @courtneymcgee, she informed me that the Hero content (e.g. header, description) will be changing infrequently. Therefore, it might make sense to serve this data from config rather than from the API.

Let me know if there was a better way to accomplish the above or any other suggestions.

### DIFF
--- a/appConfig.js
+++ b/appConfig.js
@@ -35,6 +35,15 @@ const config = {
     url: 'https://www.nypl.org/books-music-dvds/recommendations',
     label: 'Recommendations',
   },
+  listMetaData: {
+    annual: {
+      ya: {
+        category: 'Recommends',
+        header: 'Best Books for Teens',
+        description: 'Explore our annual selection of outstanding young adult titles.',
+      },
+    },
+  },
 };
 
 export default config;
--- a/appConfig.js
+++ b/appConfig.js
@@ -39,7 +39,7 @@ const config = {
     print: 'https://www.nypl.org/accessibility/print-disabilities',
     about: 'https://www.nypl.org/books-music-dvds/recommendations/about/annual-lists',
   },
-  listMetaData: {
+  heroData: {
     annual: {
       ya: {
         category: 'Recommends',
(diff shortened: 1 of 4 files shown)


test2017Data.js
```diff
@@ -3,7 +3,7 @@ export default {
   "type": "staff-picks",
   "date": "2017-01",
   "title": "Winter 2016 Staff Picks",
-  "heroImageUrl": null,
+  "heroImageUrl": 'http://staff-picks-dev.us-east-1.elasticbeanstalk.com/books-music-dvds/recommendations/staff-picks/src/client/images/desktop.banner.YA.FIN.png',
```
EdwinGuzman: This is a self-referencing image url! I know it'll get updated eventually, just wanted to point it out.

appConfig.js
```diff
@@ -35,6 +35,15 @@ const config = {
     url: 'https://www.nypl.org/books-music-dvds/recommendations',
     label: 'Recommendations',
   },
+  listMetaData: {
```
EdwinGuzman: Since we're not specifically using this for meta tag data, you can rename this to 'heroData' or maybe 'pickDescription'?

EdwinGuzman:
Very good first refactor for the header, just a minor comment about the name of the data for the hero.

Another thing, you can potentially make the `Hero` component a "pure" function. The component inherits from the React Component class but it doesn't necessarily need to. Since it's only rendering data, you can "convert" it - https://reactjs.org/docs/components-and-props.html. A recommendation but not necessary.

EdwinGuzman:
You have some minor conflicts now.

appConfig.js
kfriedman: I like `heroData`. It stay's consistent with `heroImageUrl`. I'll rename. Thanks!

PR #61, opened 2017-10-31
Using the nypl-data-api-client to make calls to the NYPL API.

EdwinGuzman:
Setting up KMS, client id/secret, and oauth endpoints. Only setting it up for DEV at the moment.

Removed the older Refinery information and added `aws-sdk` and `@nypl/nypl-data-api-client` packages.

Removed the mocked data set and am hitting `https://api.nypltech.org/api/v0.1/book-lists/kids/2017-11` directly (for the moment).

If you want to run this locally, please run:

` clientId=[xxx] clientSecret=[xxx] npm run dev-api-start`.

@kfriedman I'm using the client id/secret from Discovery, so we will need a new pair (right?).

**Note**: This is only hitting the development endpoint on the sandbox and the client id/secret and not encrypted right now.

CC: @nonword


EdwinGuzman:
Also fixes #27

PR #63, opened 2017-10-31
A11y sidebar announcement

EdwinGuzman:
Minor CSS to display it nicely next to the `Filter by Tags` heading. The aria-live is set to `assertive` for now.


EdwinGuzman:
@wlla @courtneymcgee added the `aria-label` to the filter buttons to read either the label, or "[label] remove filter". See issue #65 .

PR #82, opened 2017-11-01
Show filters only when js is enabled

### LINKED ISSUE #52: Show filters only when JS is enabled
The left sidebar filters should ONLY be displayed if JS is enabled. Another PR tracked the work to pass the boolean that checks if JS is enabled, once that is merged this task will focus on:

- Initially hiding the filters section when rendered server-side by Node
- Once the React mounts, the sidebar section will get an updated `isJsEnabled` boolean and ONLY render the filters if the value is `true`.
- A loading transition might need to be implemented -- @ricardoom or we can simply fade it in...

CC: @RobKelley

rhernand3z:
Added unit test and modified Sidebar component to only render the <BookFilters /> component if JS is enabled.

Fixes #52

### DIFF
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -48,28 +48,3 @@ body {
 li {
   list-style: none;
 }
-
-.fade{
-  opacity: 0;
-}
-
-.fade-appear-active {
-
-}
-.fade-enter {
-  opacity: 0.01;
-}
-
-.fade-enter.fade-enter-active {
-  opacity: 1;
-  transition: opacity 1000ms ease-in;
-}
-
-.fade-exit {
-  opacity: 1;
-}
-
-.fade-exit.fade-exit-active {
-  opacity: 0.01;
-  transition: opacity 800ms ease-in;
-}
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -48,3 +48,28 @@ body {
 li {
   list-style: none;
 }
+
+.fade{
+  opacity: 0;
+}
+
+.fade-appear-active {
+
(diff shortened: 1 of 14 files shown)


src/client/styles/main.scss
```diff
@@ -48,3 +48,28 @@ body {
 li {
   list-style: none;
 }
+
+.fade{
+  opacity: 0;
+}
+
+.fade-appear-active {
```
EdwinGuzman: Empty class?

src/client/styles/main.scss
```diff
@@ -48,3 +48,28 @@ body {
 li {
   list-style: none;
 }
+
+.fade{
```
EdwinGuzman: Space after the `e`.

src/client/styles/main.scss
```diff
@@ -48,3 +48,28 @@ body {
...
+}
+
+.fade-appear-active {
+
+}
+.fade-enter {
+  opacity: 0.01;
+}
+
+.fade-enter.fade-enter-active {
```
EdwinGuzman: Why not nest these rules?

src/client/styles/main.scss
rhernand3z: removed the styles for now, until we clarify the animation expectation...

src/client/styles/main.scss
rhernand3z: removed for now, I was experimenting via react-css-transition-group but don't want to go too far with it because we haven not set the expectation on if/how this should animate

EdwinGuzman:
Got it 🆗

Issue #60, opened 2017-10-31
Revise Hero content


wlla:
1) Remove the text "RECOMMENDATIONS" (`data-reactid="172"`) from the hero. 
* Note: As of 11/1, text appears as "RECOMMENDS" in the hero. We still want it removed from both kids and teens lists. 

2) Change "Best Books for Children" to "Best Books for Kids"

Issue #59, opened 2017-10-31
Adjustments to headers


wlla:
- [x] In the Hero, the text "RECOMMENDATIONS" is an `h3`. It should be an `p`. (It is being removed, but I am creating a separate ticket for that because there may be visual design implications.

- [x] In the Hero, the text "Best Books for Teens" (`data-reactid="173"`) is a `p`. It should be `h1`.

- [x] In Book cards, the titles (`class="book-item-title"`) are `h4`, they should be `h3`.

- [x] Currently, "Additional Information" (`data-reactid="3068"`) is an `h3`. Should be an `h2`. (This duplicates a comment in the PR #51.)

rhernand3z:
@wlla -- I adjusted the `book-item-title` from an `h4` to an `h3` in #69 
@ricardoom @wlla - I've also removed the `<a>` tag in the title in #69

EdwinGuzman:
Closing. All the tasks are complete.

PR #84, opened 2017-11-01
Hero accessibility

kfriedman:
This PR addresses #59 and #60. It:

- Removes the "category" heading
- Changes the main header to a `h1`.
- Updates the main header copy

### DIFF
--- a/src/app/components/Hero/Hero.jsx
+++ b/src/app/components/Hero/Hero.jsx
@@ -16,10 +16,9 @@ class Hero extends React.Component {
         <div className="Hero__container nypl-full-width-wrapper">
           <div className="Hero__text nypl-column-three-quarters nypl-column-offset-one">
             <div className="Hero__text__HeroTitle">
-              <h3>{this.props.heroData.category}</h3>
-              <p className="Hero__text__HeroTitle__des">
+              <h1 className="Hero__text__HeroTitle__des">
                 {this.props.heroData.header}
-              </p>
+              </h1>
               <p className="Hero__text__HeroTitle__intro">
                 {this.props.heroData.description}
               </p>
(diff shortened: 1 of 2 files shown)


src/app/components/Hero/Hero.jsx
```diff
@@ -16,10 +16,10 @@ class Hero extends React.Component {
         <div className="Hero__container nypl-full-width-wrapper">
           <div className="Hero__text nypl-column-three-quarters nypl-column-offset-one">
             <div className="Hero__text__HeroTitle">
-              <h3>{this.props.heroData.category}</h3>
-              <p className="Hero__text__HeroTitle__des">
+              {this.props.heroData.category ? <p>{this.props.heroData.category}</p> : ''}
```
EdwinGuzman: I think they just want to completely remove the "category" from the hero? Maybe @courtneymcgee or @wlla can clarify. If so, we wouldn't need this property in the config.

kfriedman: Um, I guess I wasn't sure if this was being removed _permanently_ (i.e. for future Staff Picks also). So I thought the best thing to do for now was to make the display determined by config. But I'm okay with removing it too!

EdwinGuzman: I honestly have no clue what the Hero content will contain. Everything else is good so let's merge and update later once we have better requirements.

wlla:
@courtneymcgee should make the final call/confirmation, but yeah _permanent_ removal (i.e. for future Staff Picks also) is what I had in mind. Context: When redesigned app is used more broadly, we'll include the full breadcrumb pattern which includes that category information. Plus, it's current placement is a wrench in clear heading organization.

PR #90, opened 2017-11-03
Update hero images

ricardoom:
for the new contents


src/app/components/Hero/Hero.jsx
```diff
@@ -10,17 +10,17 @@ class Hero extends React.Component {
...
       >
-        <div className="Hero__container nypl-full-width-wrapper">
-          <div className="Hero__text nypl-column-three-quarters nypl-column-offset-one">
-            <div className="Hero__text__HeroTitle">
+        <div className="hero__container nypl-full-width-wrapper">
+          <div className="hero__text nypl-column-three-quarters nypl-column-offset-one">
+            <div className="hero__text__HeroTitle">
               {this.props.heroData.category ? <p>{this.props.heroData.category}</p> : ''}
-              <h1 className="Hero__text__HeroTitle__des">
+              <h1 className="hero__text__HeroTitle__des">
```
EdwinGuzman: `HeroTitle` is capitalize but the next one is `heroTitle`, not sure how much the CSS relies on this.

appConfig.js
```diff
@@ -49,14 +49,14 @@ const config = {
         header: 'Best Books for Teens',
         description: 'Explore our annual selection of outstanding young adult titles.',
         heroImageUrl: 'http://staff-picks-dev.us-east-1.elasticbeanstalk.com/books-music-dvds/' +
-          'recommendations/staff-picks/src/client/images/desktop.banner.YA.FIN.png',
+          'recommendations/staff-picks/src/client/images/desktop.teens.cover.2017.png',
```
EdwinGuzman: Having or not having `staff-picks` doesn't change anything. We should pick one over the other though.

ricardoom: in the URL or more generally?

EdwinGuzman:
Oh, one of the `Hero` unit test fails. Run `npm run test` to see which one it is.

appConfig.js
EdwinGuzman: Yes, in the URL. I noticed in the image reference below, `staff-picks` isn't there, but both images load just fine.

PR #96, opened 2017-11-06
Display tags when js disabled

### LINKED ISSUE #53: Hide BookCard tags when JS is enabled
Initially the book cards should contain the tags associated with each book. Once JS is enabled, the tags should be removed.

rhernand3z:
Fixes #53 

Updated styles and component to check if JS is enabled which adds `visuallyHidden` class to the tags wrapper. Updated unit test as well.


rhernand3z:
As per my convo with @courtneymcgee -- I will merge this and then we will consider options (ex: fadeOut css animations)

CC: @ricardoom @EdwinGuzman @wlla

PR #103, opened 2017-11-08
Api kms client

EdwinGuzman:
Please check out the `kmsClientHelper.js` file. The ultimate goal is to make that file into a separate npm package.

Why? Because this same process is being used in Staff Picks and in Discovery.

Context:
We are using the [@nypl/nypl-data-api-client](https://www.npmjs.com/package/@nypl/nypl-data-api-client) and passing in each app's own client id and client secret (and oauth url) in order to make requests to the API.

The only issue is that we want to use encrypted versions of the client id and client secret as environment variables in AWS Elastic Beanstalk. In order to encrypt and decrypt, we are using AWS's KMS feature.

In AWS EB:
- Each app reads the encrypted environment variables,
- KMS decrypts the strings and instantiates the `nyplDataApiClient`,
- That client is stored and used to make requests

Locally:
- It's fine if we **don't** use encrypted environment variables. This also makes the start command that much shorter and easier in terminal,
- AWS KMS is **not** used so it's not instantiated,
- The `nyplDataApiClient` is used as before

I will write better comments and a better readme if this PR goes through in the npm package. Let me know what you think.

### DIFF
--- a/src/server/helper/kmsClientHelper.js
+++ b/src/server/helper/kmsClientHelper.js
@@ -1,10 +1,21 @@
-import NyplApiClient from '@nypl/nypl-data-api-client';
+import NyplDataApiClient from '@nypl/nypl-data-api-client';
 import aws from 'aws-sdk';
 
-const kmsClientHelper = (appEnvironment, kmsEnvironment, apiBase, tokenUrl) => {
+const kmsClientHelper = (options) => {
+  const {
+    kmsEnvironment,
+    apiBase,
+    tokenUrl,
+    clientId,
+    clientSecret,
+  } = options;
+  const keys = [clientId, clientSecret];
+  const CACHE = {};
   let decryptKMS;
   let kms;
 
+  // If we want to use encrypted client id and secret strings, then we need to set up
+  // AWS and the KMS decryption function.
   if (kmsEnvironment === 'encrypted') {
     kms = new aws.KMS({
       region: 'us-east-1',
@@ -27,22 +38,19 @@ const kmsClientHelper = (appEnvironment, kmsEnvironment, apiBase, tokenUrl) => {
     };
   }
 
-  const clientId = process.env.clientId;
-  const clientSecret = process.env.clientSecret;
-
-  const keys = [clientId, clientSecret];
-  const CACHE = {};
-
   function client() {
+    // If we already instantiated the NYPL Api Data Client, then just return it.
     if (CACHE.nyplApiClient) {
       return Promise.resolve(CACHE.nyplApiClient);
     }
 
+    // If we want to use encrypted client id and secret strings, then we first need to
+    // decrypt the strings and instantiate the NyplDataApiClient with those decrypted values.
     if (kmsEnvironment === 'encrypted') {
       return new Promise((resolve, reject) => {
         Promise.all(keys.map(decryptKMS))
           .then(([decryptedClientId, decryptedClientSecret]) => {
-            const nyplApiClient = new NyplApiClient({
+            const nyplApiClient = new NyplDataApiClient({
               base_url: apiBase,
               oauth_key: decryptedClientId,
               oauth_secret: decryptedClientSecret,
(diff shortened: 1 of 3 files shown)


src/server/helper/kmsClientHelper.js
```diff
@@ -0,0 +1,91 @@
...
+    // If we already instantiated the NYPL Api Data Client, then just return it.
+    if (CACHE.nyplApiClient) {
+      return Promise.resolve(CACHE.nyplApiClient);
+    }
+
+    // If we want to use encrypted client id and secret strings, then we first need to
+    // decrypt the strings and instantiate the NyplDataApiClient with those decrypted values.
+    if (kmsEnvironment === 'encrypted') {
+      return new Promise((resolve, reject) => {
+        Promise.all(keys.map(decryptKMS))
```
nonword: `Promise.all` returns a promise, so instead of returning a `new Promise` on line 50, I think technically you could just `return Promise.all` (and then simply `return nyplApiClient` on line 64 & `throw` the error on 68)

src/server/helper/kmsClientHelper.js
```diff
@@ -0,0 +1,91 @@
...
+    // If we are using unencrypted strings, then simply use those.
+    const nyplApiClient = new NyplDataApiClient({
+      base_url: apiBase,
+      oauth_key: clientId,
+      oauth_secret: clientSecret,
+      oauth_url: tokenUrl,
+    });
+
+    CACHE.clientId = clientId;
+    CACHE.clientSecret = clientSecret;
```
nonword: Minor point, but strictly speaking doesn't look like clientId and clientSecret need to be stored in CACHE. They're not used.

src/server/helper/kmsClientHelper.js
```diff
@@ -0,0 +1,91 @@
...
+      region: 'us-east-1',
+    });
+
+    decryptKMS = (key) => {
+      const params = {
+        CiphertextBlob: new Buffer(key, 'base64'),
+      };
+
+      return new Promise((resolve, reject) => {
+        kms.decrypt(params, (err, data) => {
```
nonword: So overall this module is a factory for making instances of NyplDataApiClient. One of the things it has to do is decrypt creds. Does the app not need to `kms.decrypt` anything else? If this is the only place you need to call `kms.decrypt`, this is fine, but maybe that could be a separate helper module like `kmsDecryptHelper` solely responsible for returning a promise that resolves decrypted text

EdwinGuzman: Very good point. Right now only the client id and secret will need to be decrypted but it'd be good to separate it out.

src/server/helper/kmsClientHelper.js
EdwinGuzman: Will remove.

src/server/helper/kmsClientHelper.js
EdwinGuzman: Thanks! I knew it looked odd and wanted to update it. Yes, it does return a promise so the outer one can be removed.

EdwinGuzman:
Updated.

nonword:
> The ultimate goal is to make that file into a separate npm package.

That could be cool. I've had similar thoughts cause it does feel like we're doing the same thing a few different palces. I've been wondering about a general purpose "nypl encrypted env reader" module that hooks into kms to automatically decrypt environmental variables based on an internal naming convention. (e.g. Any ENV variable ending in "_ENCRYPTED" is assumed to be encrypted) So, for example, I could instantiate a NyplDataApiClient via:

```
require('@nypl/nypl-encrypted-env-reader')
  .load()
  .then((decryptedEnv) => {
    // create api client with decrypted env variables:
    let apiClient = new require('@nypl/nypl-data-api-client')({
       base_url: decryptedEnv.NYPL_API_BASE_URL,
       oauth_key: decryptedEnv.NYPL_OAUTH_KEY,
       oauth_secret: decryptedEnv.NYPL_OAUTH_SECRET,
       oauth_url: decryptedEnv.NYPL_OAUTH_URL,
    })

    // create google drive client with decrypted env variables:
    let googleDriveClient = new GoogleDriveClient({
      private_key: decryptedEnv.GOOGLE_DRIVE_PRIVATE_KEY
    })
  })
```

By keeping NyplDataApiClient concerns *outside* the general use module, the module could be used for other decrypted config. This is beyond the scope of this PR, but interested to discuss later

EdwinGuzman:
That's a fantastic idea. I'll merge this in and work on a further step to do what you suggested and share it with you.

PR #122, opened 2017-11-14
Removing aria-labelledby and using aria-label for the filter list.

### LINKED ISSUE #115: Duplicated tag list label
We now have a label associated with the list of tags, however, the label is formatted as a visually hidden span which is then referenced by the label. This causes the text to be repeated when moving down the page sequentially. Review usability.

EdwinGuzman:
Fixes #115 

I removed the span and used the `aria-label` in the `ul` itself. Another option that @wlla suggested was using the h2 'Filter by Tags' label as the label for the `ul`. Which do you think is better, @courtneymcgee ?

This is up on prod: http://staff-picks-production.us-east-1.elasticbeanstalk.com/books-music-dvds/recommendations/best-books/ya

### DIFF
--- a/src/app/components/BookFilters/BookFilters.jsx
+++ b/src/app/components/BookFilters/BookFilters.jsx
@@ -139,10 +139,7 @@ class BookFilters extends React.Component {
           </button>
         </div>
         <div className={`book-filters-list ${filtersContainerDisplayClass}`}>
-          <span className="visuallyHidden" id="filter-list-description">
-            Click to apply or remove tags.
-          </span>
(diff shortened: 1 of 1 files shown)


EdwinGuzman:
Which approach?

courtneymcgee:
The one you took here that I approved. Am I missing something?

EdwinGuzman:
Oh yeah, I wrote what Willa suggested and asked which you thought was better. I'm just going to go with what's in the PR.

courtneymcgee:
Great!

PR #123, opened 2017-11-14
Updating the clear filters button for tablet view.

### LINKED ISSUE #99: Clear Filters doesn’t look like a button on phone/tablet viewports
@chrismulholland picked this up in testing. I've taken a look and it does look pretty stretched / non-button-y at various mobile sizes. @ricardoom, can you check it out and make a recommendation?

EdwinGuzman:
Fixes #99 

Check it out on dev for tablet view.

### DIFF
--- a/src/client/styles/components/_bookFilters.scss
+++ b/src/client/styles/components/_bookFilters.scss
@@ -103,12 +103,12 @@
   .clear-button {
     width: 88%;
     text-align: right;
-    border-radius: 5px;
     font-size: 0.8rem;
     padding: 10px;
     font-size: 1rem;
     margin: 0 0 0 1.3em;
     width: 200px;
+    @include border-radius(5px);
 
     @include min-screen($mobile-breakpoint) {
       width: 100%;
--- a/src/client/styles/components/_bookFilters.scss
+++ b/src/client/styles/components/_bookFilters.scss
@@ -103,11 +103,17 @@
   .clear-button {
     width: 88%;
     text-align: right;
-    border-radius: 0px;
+    border-radius: 5px;
     font-size: 0.8rem;
     padding: 10px;
     font-size: 1rem;
     margin: 0 0 0 1.3em;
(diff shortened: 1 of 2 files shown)


src/client/styles/components/_bookFilters.scss
```diff
@@ -103,11 +103,17 @@
   .clear-button {
     width: 88%;
     text-align: right;
-    border-radius: 0px;
+    border-radius: 5px;
```
rhernand3z: there is a mixin for this to tackle cross-browser compatibility...

rhernand3z:
When running locally, I see the following issues on both FF and Chrome:

1) When the tag text is long, the X is right next to it...
screen shot 2017-11-14 at 3 38 02 pm

2) On viewport 965px width, it does not adapt to a 2 column layout. I see a huge white space on the right (see screenshot)
screen shot 2017-11-14 at 3 38 25 pm

EdwinGuzman:
Right, the breakpoint is exactly 965px, and we're not doing "[width size] + 1", so that only happens for that exact width.

rhernand3z:
Shouldn't the filters auto fit into columns up until the grid splits into 1/3 | 3/4 ? I assume we should account up until that split happens?

chrismulholland:
image

EdwinGuzman:
Cool. The 'X' being too close to the filter text is not directly related to this issue so I'm going to create another issue. Thanks.

chrismulholland:
is the multi-column button layout supposed to appear in all browsers at certain widths?  If so, I don't see it in Chrome.  Firefox, yes.

EdwinGuzman:
It's suppose to appear in viewports between 483px and 966px.

EdwinGuzman:
@rhernand3z @chrismulholland check dev now. It should be multi-column at 965px + 1 (which I'm not liking too much) and the padding should be better between the X and text.

Issue #117, opened 2017-11-13
Aria-live issues with Jaws/Chrome


wlla:
The total books found `aria-live` region has a delay when using Jaws and Chrome on a PC. The number announced is the number for the region version immediately preceding. Functionality is working correctly in Jaws/IE/Firefox, VoiceOver/Safari, and mobile (Android + iOS).

EdwinGuzman:
Could be related to #114. Check out the resource. It could be a specific version and browser that's the issue we cannot readily solve.

seanredmond:
I see the same thing happening with Chrome and VoiceOver. I looks to me like VoiceOver is quicker on the draw than the DOM changes?

Issue #137, opened 2017-11-16
First filter button often gets "X books found" text in screenreader


seanredmond:
When tabbing with Chrome and VoiceOver on from the "X books found" message to the first filter button, the text of the button is is not read. Instead the "X books found" is read instead. If you are shift-tabbing in the other direction, the first filter label is read fine.

Watching what's displayed on the screen by VoiceOver, I can see that , for a split second, it's about to say the button label but that it gets overriden by the books found. I wonder if focusing the button is triggering an event that make VoiceOver think there has been a change that aria-live should announce even thought he change is from 50 books to 50 books? 

Not sure if this is related to #118 or not.

courtneymcgee:
@wlla , @seanredmond references #118 (Inaccurate list of tags displayed with Chrome and Jaws) above, but might this be a result of #105 (Once a filter is applied, focus should go to top of list), which we put in place to address #101 (Aria-live region is not announcing the change in the number of books)? Apologies for all of the references! I'd love your opinion, as I'm trying to make a call on if we can or will address.

PR #138, opened 2017-11-16
Removing aria-live from filter list and delaying the focus to the # books found.

EdwinGuzman:
Testing removing `aria-live` and adding a short delay to the focus for better screenreader interaction.


EdwinGuzman:
Thanks. Added a cleartimeout and set the timeout to be stored in the state.

EdwinGuzman:
@courtneymcgee gave the 👍 

Will merge soon.

EdwinGuzman:
Fixes #117 and #137

PR #147, opened 2017-12-05
Removing visually hidden css from the app.

EdwinGuzman:
This PR uses the same header we have but removes the `.visuallyHidden` css rule from the app and relies on the one set from Design Toolkit. I'm not sure why we have two different sets but we should just use the DT rule.

This solves the problem of the NYPL Logo and the Locations link not being tappable on VO on mobile devices. The only issue is how the VO focus wraps the element. It should be address but I'm not sure how to start that and I think @ricardoom mentioned it could be the SVG box (?).

Related Header tickets: [HDR-21](https://jira.nypl.org/browse/HDR-21) [HDR-12](https://jira.nypl.org/browse/HDR-12)

Currently up on dev: http://staff-picks-dev.us-east-1.elasticbeanstalk.com/books-music-dvds/recommendations/best-books/ya

image uploaded from ios 5

### DIFF
--- a/src/app/components/BookFilters/Filter.jsx
+++ b/src/app/components/BookFilters/Filter.jsx
@@ -112,6 +112,7 @@ Filter.propTypes = {
 
 Filter.defaultProps = {
   clearTimeout: () => {},
+  setDisabled: () => {},
 };
 
 export default Filter;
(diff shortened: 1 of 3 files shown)


rhernand3z:
The reason we have two was because no one looked at the current establishment set prior to the toolkit. The toolkit is relatively new and most of our React Apps had integrated this _mixins.scss file in the past. The solution would be to simply rely on the toolkit for all these global utility classes and remove all references in the apps (which you are already doing).

Not sure I am 100% convinced of relying on the toolkit for this because at the moment IMO the toolkit has very verbose logic in there and NEEDS to be cleaned up.

PR #150, opened 2018-01-09
Setup Travis Continuous Intergration and Delivery to Elastic Beanstalk on AWS

rhernand3z:
This PR implements a CI/CD pipeline with Travis. Utilizing travis.yml to kick off the `deploy` phase only on the `master` branch.

AWS configurations such as `access_key_id`, `access_secret`, `aws_application_name` and `aws_environment_name` have been setup as Travis `ENV` variables in the repository settings.

The `access_key_id` AND `access_secret` ENV variables have been encrypted in the repo settings page, those credentials are assigned to the user `TravisElasticBeanstalk` on our AWS Digital Dev account which contains the proper policy for triggering changes to Elastic Beanstalk.

Ideally, the `travis.yml` has been setup so it can be easily configured on an app by app basis in the Travis repo settings UI.

If the `before_script` task or `script` task fails, the deploy will NOT be triggered. This is essentially what we want in a full CI/CD pipeline where automation of deployments only occur on successful execution of test coverage.

This will improve our AWS EB workflow and will remove the manual deploy triggers.


.travis.yml
```diff
@@ -1,6 +1,24 @@
...
+    - node_modules
 install: npm install
 before_script: npm run dist
 script: npm test
+before_deploy: "echo 'All unit tests passed; Preparing to deploy Staff Picks to AWS'"
+deploy:
+  skip_cleanup: true
+  provider: elasticbeanstalk
+  access_key_id: $AWS_PRODUCTION_ACCESS_KEY_ID
+  secret_access_key: "$AWS_PRODUCTION_ACCESS_SECRET"
```
rhernand3z: necessary to wrap secret in quotes to handle encrypted string with special chars.

nonword: Is this encrypted? Do you not need to add a `secure:` sub-property per https://docs.travis-ci.com/user/encryption-keys#Usage ?

rhernand3z: Yes it is encrypted by the Travis repo settings.

Take a look at the log, it displays [secure] -- https://travis-ci.org/NYPL/staff-picks/builds/326922081#L429

Documentation: https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

> By default, the value of these new environment variables is hidden from the export line in the logs. This corresponds to the behavior of encrypted variables in your .travis.yml. The variables are stored encrypted in our systems, and get decrypted when the build script is generated.

nonword: I see. But do you know how travis *knows* they're encrypted if they're not marked as such in `.travis.yml`? How does travis know which env vars need to be decrypted before use, in other words?

rhernand3z: If you set `ENV` variables in the repository settings and you do not select `Display value in build log` -- users do not see the value in the UI, in addition Travis stores them encrypted and decrypts when the build script is generated.

If you take a look at this build in Travis (https://github.com/NYPL/staff-picks/commit/3f76e5c7f55f29615dc2418ac6e11705c5aea40c) -- I added the `secure` sub-property and attempted to use an encrypted key, it failed every time. In my search for an answer, it seems it has something to do when passing encrypted strings that contain special characters (specifically the forward/backslash). 

Although the `.travis.yml` file is not explicitly stating the `secure` sub-property. The Travis repository settings is handling the decryption/encryption of these values and I don't see where they become visible or can be a security risk?

kfriedman: I _think_ in this case, Travis is getting the value from the [Repository Setting](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) so it's not necessary to use the `secure` syntax. Right?

You _would_ have to use the `secure` syntax if you were saving the encrypted variable in `travis.yml` (and in source control) and, as Paul, mentioned Travis would need to know that the variable is encrypted.

rhernand3z: Yes that is correct, I have a commit where I used it successfully.. However I did not want to expose any credentials in the yml file and wanted to ensure it can be easily swapped by the settings section.

PR #151, opened 2018-02-07
Setup travis ci aws,  swap-space config and update documentation

rhernand3z:
This PR completes the process of integrating Travis for CI/CD on AWS. In addition, I've added the swap-space config to address the issues we have had in the past. Once merged, travis should auto deploy.


.travis.yml
```diff
@@ -1,24 +1,52 @@
...
+- provider: elasticbeanstalk
+  skip_cleanup: true
+  access_key_id: "$aws_access_key_id_development"
+  secret_access_key: "$aws_secret_access_key_development"
+  region: us-east-1
+  app: "$aws_app_name"
+  env: "$aws_environment_name_development"
+  bucket_name: elasticbeanstalk-us-east-1-224280085904
+  bucket_path: "$aws_app_name"
+  on:
```
EdwinGuzman: You have this above with `branch: master`. Is it okay to two of these??

rhernand3z: Since QA and Prod environments both live in digital-dev account, they use the same credentials.

PR #154, opened 2018-02-09
More EB refinements: HTTPS and managed updates

kfriedman:
Hi all!

This PR adds a couple of more (and, hopefully, last) refinements to our EB setup:

- Enabling HTTPS on the load balancer
- Enabling [managed updates](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environment-platform-update-managed.html) and setting an update window for EB instances

Managed updates might be a bit more controversial as we've never done that before but I think it's a good idea for security and maintenance reasons.

Also, I've spoke with @rhernand3z about this but I'm not 100% sure how to handle the HTTPS setup in our two environments (i.e. `nypl-sandbox` and `nypl-digital-dev`) since the ARN of the SSL certificate will be different. For now, I've just included them both as "sample files" that you can then copy and rename to the `.ebextensions` directory.

Let me know your thoughts!


EdwinGuzman:
Great. So, for now, we should copy and paste the `SSLCertificateID` that's associated to the location we want to deploy to (either sandbox or digital-dev)? That seems fine for me but easy to forget. Still, a good step forward.

kfriedman:
I guess the thinking was you'd just copy the _entire_ appropriate file from the `https_configs` directory depending on the AWS account. Perhaps someone smarter than me can come up with a better solution.

EdwinGuzman:
Added @holingpoon  for another set of eyes and to merge this in with other Travis related work.

holingpoon:
This is interesting. I was talking to @nodanaonlyzuul about this where the difference is in the `SSLCertificateId`. I'm approving this since we should track this in the code base. I'm digging around documentation and forums to see if we can do environment detection and switch between the two `SSLCertificateId`s.

PR #158, opened 2018-04-06
Refactor update

EdwinGuzman:
Comments here and there.
Updated eslint, added some rules, and fixed issues.


src/server/helper/kmsClientHelper.js
```diff
@@ -60,7 +60,7 @@ const kmsClientHelper = (options) => {
 
           return nyplApiClient;
         })
-        .catch(error => {
+        .catch((error) => {
```
ktp242: I think if only we have more than one arguments then we need `()`?

PR #159, opened 2018-04-09
Kp/www 72/create form for date input

ktp242:
This PR focus on creating the basic form of creates  season dropdown menu. It adds a new component for the menu. It also adds some dummy data for running the apps and generate the dropdown menu.

### DIFF
--- a/appConfig.js
+++ b/appConfig.js
@@ -89,6 +89,12 @@ const config = {
   loginUrl: 'https://login.nypl.org/auth/login',
   tokenUrl: 'https://isso.nypl.org/',
   heroData: {
+    staffPicks: {
+      category: '',
+      header: 'Staff Picks',
+      description: 'This is staff picks page.',
+      heroImageUrl: 'src/client/images/desktop.teens.cover.2017.png',
+    },
     annual: {
       ya: {
         category: '',
--- a/src/app/components/Application/Application.jsx
+++ b/src/app/components/Application/Application.jsx
@@ -37,6 +37,7 @@ class App extends React.Component {
   }
 
   render() {
+    // temporarily add the check here for staff picks config
     let heroData = undefined;
 
     if (_isEmpty(this.props.params)) {
--- a/src/app/components/Application/Application.jsx
+++ b/src/app/components/Application/Application.jsx
@@ -11,6 +11,8 @@ import BookActions from '../../actions/BookActions';
 import BookStore from '../../stores/BookStore';
 import config from '../../../../appConfig.js';
 
+import { isEmpty as _isEmpty } from 'underscore';
+
 class App extends React.Component {
   constructor(props) {
     super(props);
@@ -35,6 +37,14 @@ class App extends React.Component {
   }
 
   render() {
+    let heroData = undefined;
+
+    if (_isEmpty(this.props.params)) {
+      heroData = config.heroData.staffPicks;
+    } else {
+      heroData = config.heroData.annual[this.props.params.type];
+    }
+
     return (
       <div className="app-wrapper">
         <Header
@@ -44,7 +54,7 @@ class App extends React.Component {
 
         <main className="main-page">
           <Hero
-            heroData={config.heroData.annual[this.props.params.type]}
+            heroData={heroData}
           />
 
           <div id="app-content" className="nypl-full-width-wrapper">
--- a/src/app/components/ListFilter/ListFilter.jsx
+++ b/src/app/components/ListFilter/ListFilter.jsx
@@ -36,10 +36,11 @@ class ListFilter extends React.Component {
     const visuallyHidden = (this.state.withJS) ? 'visuallyHidden' : '';
     const selectName = fieldsetProps.fieldsetName;
     const selectId = `${selectName}-input`;
-    const defaultValue = fieldsetProps.options[0].value;
-    const optionList = fieldsetProps.options.map(
+    const defaultValue = (fieldsetProps.options.length && fieldsetProps.options[0].value) ?
+      fieldsetProps.options[0].value : '';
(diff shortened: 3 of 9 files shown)


src/app/components/Application/Application.jsx
```diff
@@ -35,6 +37,15 @@ class App extends React.Component {
   }
 
   render() {
+    // temporarily add the check here for staff picks config
+    let heroData = undefined;
+
+    if (_isEmpty(this.props.params)) {
```
ktp242: So is the condition check here for temporarily running staff picks route

src/app/components/Sidebar/Sidebar.jsx
```diff
@@ -3,8 +3,20 @@ import PropTypes from 'prop-types';
 import { LeftWedgeIcon } from '@nypl/dgx-svg-icons';
 
 import BookFilters from '../BookFilters/BookFilters.jsx';
+import ListFilter from '../ListFilter/ListFilter.jsx';
 import config from '../../../../appConfig';
 
+// This data is a temporary dummy data for creating the season list
+const fieldsetProps = {
```
ktp242: The dummy data of the season list. It will come from the API (I assume) in the future PRs.

src/app/components/ListFilter/ListFilter.jsx
```diff
@@ -0,0 +1,77 @@
...
+    this.state = {
+      withJS: false,
+      submitValue: '2018-01-01',
+    };
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  componentDidMount() {
+    this.setState({
```
EdwinGuzman: There's an `action` that does the same thing at the top `Application` level. That prop `isJsEnabled` is passed down to all the children components so use that instead of setting the state for in this component.

src/app/components/ListFilter/ListFilter.jsx
```diff
@@ -0,0 +1,77 @@
...
+      }
+    );
+  }
+
+  renderFieldset() {
+    const fieldsetProps = this.props.fieldsetProps;
+    const visuallyHidden = (this.state.withJS) ? 'visuallyHidden' : '';
+    const selectName = fieldsetProps.fieldsetName;
+    const selectId = `${selectName}-input`;
+    const defaultValue = (fieldsetProps.options.length && fieldsetProps.options[0].value) ?
```
EdwinGuzman: This is fine for now, but if you go to `Spring 2017` directly that won't be the first object in the array so you can't always do `options[0]`.

src/app/components/ListFilter/ListFilter.jsx
```diff
@@ -0,0 +1,77 @@
...
+      fieldsetProps.options[0].value : '';
+    const optionList = (fieldsetProps.options.length) ? fieldsetProps.options.map(
+      (opt) => <option value={opt.value} key={opt.value}>{opt.name}</option>
+    ) : null;
+
+    return (
+      <fieldset>
```
EdwinGuzman: I think this fieldset should be its own component rather than being part of `ListFilter`. That way, you can reuse that component for the `date` and for the `audience` data.

EdwinGuzman:
It works well without JS too. It's not hooked up yet but I noticed that it adds the option selected as a query param in the URL. We can fix that later once we have the server functions that call the API.

appConfig.js
```diff
@@ -89,6 +89,12 @@ const config = {
   loginUrl: 'https://login.nypl.org/auth/login',
   tokenUrl: 'https://isso.nypl.org/',
   heroData: {
+    staffPicks: {
```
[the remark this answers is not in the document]
ktp242: /books-music-dvds/recommendations/staff-picks/

src/app/components/ListFilter/ListFilter.jsx
ktp242: My plan is to use <ListFilter> for both `date` and `audience`. Because `date` and `audience` are independent two forms.

src/app/components/ListFilter/ListFilter.jsx
ktp242: OK we can use that then!

src/app/components/ListFilter/ListFilter.jsx
ktp242: I see your point. I think `date` and `audience` should be two forms with their own submit button, yet you think they should be one form together with one submit button. I think with JS it won't make difference. But without JS, one form probably better so the user won't need to click submit twice if they want to change both audience and date.

EdwinGuzman: Correct. One form will be better because you'll only have to have one submit button. That way the user can select "Spring 2016" and "Children" and get the correct form without having to submit "Spring 2016" first and then wait to submit "Children" after.

It doesn't make too much of a difference with JS but you do want to have semantic DOM elements and it's better to have two fieldsets in one form than two forms.

test/unit/ListSelector.test.js
```diff
@@ -0,0 +1,43 @@
...
+    fieldsetName: 'audience',
+    options: [
+      { name: 'Adult', value: 'adult' },
+      { name: 'Teen', value: 'teen' },
+      { name: 'Children', value: 'children' },
+    ],
+  },
+};
+
+describe('ListSelector', () => {
```
EdwinGuzman: You could use a few more tests to make sure two `select` inputs are rendered.

EdwinGuzman:
Looks good to me. I'd say see if you can make the `ListFilter` component more generic so that it can be called something like `SelectInput` and be used in other projects. The more modular, the better, but not necessary. So approving.

PR #160, opened 2018-04-13
Kp/www 73/making api call to dev book lists

ktp242:
This PR focuses on getting the data from /book-lists API from the client and representing the books. For doing this, it adjusted the order of the routes on the server side. It also includes related tests.

Another update is to use `.env` for inject environment variables, so we don't have to include it every time when running the application from terminal. Now simply run `npm run dev-api-start`. Related documentation for doing that is updated at README


src/app/components/Sidebar/Sidebar.jsx
```diff
@@ -11,17 +11,17 @@ const fieldsetProps = {
...
     fieldsetName: 'season',
     options: [
-      { name: '2018 Winter', value: '2018-01-01' },
-      { name: '2017 Fall', value: '2017-09-01' },
-      { name: '2017 Summer', value: '2017-06-01' },
-      { name: '2017 Spring', value: '2017-04-01' },
+      { name: '2018 Winter', value: '2018-01' },
+      { name: '2017 Fall', value: '2017-09' },
+      { name: '2017 Summer', value: '2017-06' },
+      { name: '2017 Spring', value: '2017-04' },
```
gkallenberg: Will we be dynamically generating this list from the `title` and list's `slug`?

src/server/routes/monthData.js
```diff
@@ -28,7 +33,35 @@ function selectMonthData(req, res, next) {
...
+        statusCode: error.statusCode,
+        errorMessage: error.code,
+      });
+    });
+}
+
 export default {
   currentMonthData,
   selectMonthData,
+  selectClientMonthData,
```
gkallenberg: Just curious here, but why the name change to selectClientMonthData? Also, don't we need to make a route change for lists so that /staff-picks/2017-01 still provides the requested list? Individual book pages are going to need /staff-picks/2017-01/[isbn-title] routes in order to provide backwards compatibility, so perhaps the list routes also need to exist for the same.

When the selector is changed, the list updates just fine which is great, but the route does not include the URL param for its identifier `2017-01`. Am I incorrect in my thinking?

platformConfig.js
```diff
@@ -0,0 +1,19 @@
...
+  appEnv: process.env.APP_ENV,
+  kmsEnvironment: process.env.KMS_ENV,
+  api: {
+    development: 'https://dev-platform.nypl.org/api/v0.1',
+    qa: 'https://qa-platform.nypl.org/api/v0.1',
+    production: 'https://platform.nypl.org/api/v0.1',
+  },
+};
+
+export default platformConfig;
```
gkallenberg: Excellent to break out some configs to a new file since that config file is really long.

src/app/components/Sidebar/Sidebar.jsx
ktp242: They are just temporary.  I will switch it to come from the endpoint once we have more data.

readme.md
```diff
@@ -36,14 +46,15 @@ $ clientId=[clientId] clientSecret=[clientSecret] npm run dev-api-start
...
+$ NODE_ENV=production npm run dev-api-start
 ```
 
 To use the `production` API, run:
 ```sh
-$ NODE_ENV=production clientId=[clientId] clientSecret=[clientSecret] npm run prod-api-start
+$ NODE_ENV=production npm run prod-api-start
 ```
 
+Notice that just as running in development mode, you can still overwrite the credential settings in `.env` by adding `CLIENT_ID=[clientId] CLIENT_SECRET=[clientSecret]` to your production mode command line.
```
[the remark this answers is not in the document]
ktp242: Thanks for the review above for README.md! I will update them.

platformConfig.js
ktp242: :+1:

src/server/routes/monthData.js
ktp242: The name here is because it  updates month data but not like `selectMonthData`, it handles the requests from client side. I am actually open to a better name. Just some apps we named these two kind of functions that way.

And the second question is yes we should change the URL. I will do it the next ticket with handling server side API call.

PR #162, opened 2018-04-17
Kp/www 76 get and update url for server side request api request

ktp242:
This PR makes server side requests to `/book-lists` endpoints, so now the application can handling direct URL inputs and no JS form submissions.

It also now be able to push history for client side requests.

### DIFF
--- a/src/app/components/ListSelector/ListSelector.jsx
+++ b/src/app/components/ListSelector/ListSelector.jsx
@@ -4,10 +4,6 @@ import axios from 'axios';
 import ListFilter from './ListFilter.jsx';
 import config from '../../../../appConfig';
 import BookActions from '../../../app/actions/BookActions.js';
-import Utils from '../../utils/utils';
-
-// The module to update the URL and history with client side requests
-const history = Utils.createAppHistory();
 
 class ListSelector extends React.Component {
   constructor(props) {
@@ -18,12 +14,12 @@ class ListSelector extends React.Component {
   }
 
   /**
-   * updateHistory(url)
-   * Updates the browsing history with the URL that the client side request should go to
+   * updateLocation(url)
+   * Pushes a new location with the URL that the client side request should go to
    * @param {string} url
    */
-  updateHistory(url) {
-    history.push({ pathname: url });
+  updateLocation(url) {
+    this.props.router.push({ pathname: url });
   }
 
   /**
@@ -55,12 +51,12 @@ class ListSelector extends React.Component {
             `API error with status code ${response.data.statusCode}: ${response.data.errorMessage}`
           );
           // Lead the user to the 404 page
-          this.updateHistory('/books-music-dvds/recommendations/staff-picks/404');
+          this.updateLocation('/books-music-dvds/recommendations/staff-picks/404');
         } else {
           // For valid API response, update BookStore for the new list
           this.updateBookStore(response.data.currentPicks);
           // Update and transit to the match URL
-          this.updateHistory(
+          this.updateLocation(
             `/books-music-dvds/recommendations/staff-picks/${submitValue}`
           );
         }
@@ -78,7 +74,7 @@ class ListSelector extends React.Component {
           `${errorStatusText}`
         );
         // Lead the user to the 404 page
-        this.updateHistory('/books-music-dvds/recommendations/staff-picks/404');
+        this.updateLocation('/books-music-dvds/recommendations/staff-picks/404');
       });
   }
 
@@ -122,6 +118,7 @@ class ListSelector extends React.Component {
 ListSelector.propTypes = {
   fieldsetProps: PropTypes.object,
   isJsEnabled: PropTypes.bool,
+  router: PropTypes.object,
 };
 
 ListSelector.defaultProps = {
--- a/src/app/components/ListSelector/ListSelector.jsx
+++ b/src/app/components/ListSelector/ListSelector.jsx
@@ -4,6 +4,26 @@ import axios from 'axios';
 import ListFilter from './ListFilter.jsx';
 import config from '../../../../appConfig';
 import BookActions from '../../../app/actions/BookActions.js';
+import {
+  createHistory,
+  useQueries,
+  createMemoryHistory,
+} from 'history';
+
+/**
+ * createAppHistory
+ * Creates createHistory instance that supports both server side and client side rendering
+ */
+const createAppHistory = () => {
+  if (typeof(window) !== 'undefined') {
+    return useQueries(createHistory)();
+  }
+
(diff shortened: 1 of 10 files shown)


ktp242:
Can be tested here!
http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-dvds/recommendations/staff-picks/children-01/

It should work both with/without JS. One problem here is that if switch among the options between the audience lists, the URL will be incorrectly presented, such as `ya-01`. The issue will be solved in the next PR.

src/app/components/ListSelector/ListSelector.jsx
```diff
@@ -4,17 +4,44 @@ import axios from 'axios';
 import ListFilter from './ListFilter.jsx';
 import config from '../../../../appConfig';
 import BookActions from '../../../app/actions/BookActions.js';
+import {
```
EdwinGuzman: The whole app should know about the history and now just this component. This should go in utils somewhere or at the top level in `Application.jsx`.

src/app/components/ListSelector/ListSelector.jsx
```diff
@@ -29,22 +56,29 @@ class ListSelector extends React.Component {
...
+    axios.get(`${config.baseApiUrl}${submitValue}`)
       .then(response => {
         // Catch the error from API, and update BookStore back to the default
         if (response.data.statusCode >= 400) {
           this.updateBookStore();
           console.log(
             `API error with status code ${response.data.statusCode}: ${response.data.errorMessage}`
           );
+          // Lead the user to the 404 page
+          this.updateHistory('/books-music-dvds/recommendations/staff-picks/404');
```
EdwinGuzman: Instead of updating the history, you should use the router to go to that 404 page.

src/server/routes/monthData.js
```diff
@@ -54,14 +99,35 @@ function selectClientMonthData(req, res) {
...
       });
     });
 }
 
+/**
+ * selectClientMonthDataPost(req, res, next)
+ * Handles the requests from the form submit button (when no JS).
+ * It redirects to the route to execute the function for server side requesting.
+ */
+function selectClientMonthDataPost(req, res) {
```
EdwinGuzman: I think you mean `Server` instead of `Client`?

src/server/routes/monthData.js
```diff
@@ -54,14 +99,35 @@ function selectClientMonthData(req, res) {
...
+  const season = (req.body.season) ? `${req.body.season}-01` : '';
+  const audience = req.body.audience;
+
+  if (!season || !audience) {
+    console.log(
```
EdwinGuzman: You should also redirect to the 404 page on the server side if the from submission doesn't work.

src/server/routes/monthData.js
```diff
@@ -54,14 +99,35 @@ function selectClientMonthData(req, res) {
...
     });
 }
 
+/**
+ * selectClientMonthDataPost(req, res, next)
+ * Handles the requests from the form submit button (when no JS).
+ * It redirects to the route to execute the function for server side requesting.
+ */
+function selectClientMonthDataPost(req, res) {
+  const season = (req.body.season) ? `${req.body.season}-01` : '';
```
EdwinGuzman: I don't like how `-01` is hardcoded here. It seems likely to fail if the API is updated.

EdwinGuzman:
I noticed from the previous PR that selecting an audience changes the URL (like `staff-picks/children-01/`. That is not something we talked about doing. Those are filters for "Staff Picks" lists but the URL is `childrens` and `ya` for the "Best Books" lists.

ktp242:
The issue `children-01` has been addressed on previous comment from me. It will be solved the next PR.

src/server/routes/monthData.js
ktp242: Because from the URL, the season/month's convention always has `[year]-[month]-01` in the end. However, the API only need `[year]-[month]`. I think `-01` is redundant, but we already have it from beginning.

src/server/routes/monthData.js
ktp242: Will do it on the next PR.

src/server/routes/monthData.js
ktp242: I use `Client` because the form is from client, but yeah the real call is made by server side.

EdwinGuzman: The form submission is a POST so it's a server side since it refreshes the page. When we do ajax calls, those are client side since we request a specific endpoint to just get json back. That name sounds better.

src/server/routes/monthData.js
EdwinGuzman: Do it for this one to make it complete. It isn't hard to do.

src/server/routes/monthData.js
EdwinGuzman: I'm not sure if @gkallenberg finished working on the API yet but the previous API endpoints had `[year]-[month]-01` as the convention. If we move away from that convention, then yes it is redundant and we don't need it. But also remember that users will have URLs that are in the form of `[year]-[month]-01`.

EdwinGuzman:
Don't leave fixes for the next PR. If it's going into development then it should be working and we should be happy with it. We shouldn't merge something that is not working properly and then fix it in a later PR, unless if it's very necessary. I rather you take the time to fix it in this PR than later.

EdwinGuzman:
@ktp242  please add @holingpoon  to your PRs since she's also working on this and needs to keep updated with changes.

src/server/routes/monthData.js
ktp242: So far if you check swagger it is `[year]-[month]`, so I just follow that convention.

src/server/routes/monthData.js
ktp242: I hear you. It is not because hard or easy. I have the plan to do error handlings in another ticket so I can track it easier.

src/server/routes/monthData.js
EdwinGuzman: Greg posted that it's now `[year]-[month]-01` in Slack a few hours ago. Let's stick to that.

ktp242: Cool will update it

ktp242:
https://github.com/NYPL/staff-picks/pull/163
should fix the `children-01` issue

EdwinGuzman:
Instead of passing down the router as a prop, use the router context which is already defined in https://github.com/NYPL/staff-picks/blob/master/src/app/components/Application/Application.jsx#L68

That way, you do not have to explicitly pass it down as a prop. And, instead of doing `this.props.router` you can use it like `this.context.router`.

ktp242:
Updated to use router as context prop and removed unnecessary props.

PR #173, opened 2018-04-26
Kp/www 209 get all the books count

ktp242:
This PR display the picks and the count of the picks based on the selected audience. If a filter is selected/unselected, it will filter/cancel to filter from the audience picks.


EdwinGuzman:
Ah. This is something I mentioned I updated when I told you about the title heading component. Some updates are good though.

EdwinGuzman:
Check #174 for the changes I mean.

PR #174, opened 2018-04-26
Update dynamic tweaks

EdwinGuzman:
* Updating action call and store data structure for the app.
* Updating a few function names.
* Updating routing for the individual page.
* Updating variable names for fetching and returning data.
* Filtering picks by audience and by tag.

Initial PR but need to add tests.

### DIFF
--- a/src/app/actions/BookActions.js
+++ b/src/app/actions/BookActions.js
@@ -2,11 +2,7 @@
 import alt from '../alt';
 
 class Actions {
-  updateListType(listType) {
-    return listType;
-  }
-
-  updatePicks(picks) {
+  updatePicksData(picks) {
     return picks;
   }
 
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -1,3 +1,4 @@
+/* globals document */
 import React from 'react';
 import PropTypes from 'prop-types';
 
@@ -26,6 +27,7 @@ class Main extends React.Component {
     this.getPicksInfo = this.getPicksInfo.bind(this);
     this.getNewPickSet = this.getNewPickSet.bind(this);
     this.filterByAudience = this.filterByAudience.bind(this);
+    this.getCount = this.getCount.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -68,7 +70,13 @@ class Main extends React.Component {
     });
   }
 
-  getPicksInfo(picksData) {
+  /**
+   * getPicksInfo(picksData, currentAudience)
+   * Gets display information for a specific pick data set.
+   * @param {object} picksData
+   * @param {string} currentAudience
+   */
+  getPicksInfo(picksData, currentAudience) {
     if (!picksData) {
       return {};
     }
@@ -78,7 +86,7 @@ class Main extends React.Component {
 
     return {
       displayDate,
-      displayAge: appConfig.audienceMap[this.props.currentAudience || 'Adult'],
+      displayAge: appConfig.audienceMap[currentAudience || 'Adult'],
     };
   }
 
@@ -140,9 +148,11 @@ class Main extends React.Component {
       selectedFilters,
     });
 
-    setTimeout(() => {
-      document.getElementById('list-title').focus();
-    }, 400);
+    if (document.getElementById('list-title')) {
+      setTimeout(() => {
+        document.getElementById('list-title').focus();
+      }, 400);
+    }
   }
 
   /**
@@ -194,7 +204,7 @@ class Main extends React.Component {
           picks={this.state.picks}
           isJsEnabled={this.props.isJsEnabled}
           type={type}
-          displayInfo={this.getPicksInfo(this.props.picksData)}
+          displayInfo={this.getPicksInfo(this.props.picksData, this.props.currentAudience)}
           picksCount={picksCount}
         />
       </div>
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -1,7 +1,5 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
-import { isEmpty as _isEmpty } from 'underscore';
 
 import BookList from '../BookList/BookList';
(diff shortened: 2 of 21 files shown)


src/app/actions/BookActions.js
```diff
@@ -2,11 +2,7 @@
 import alt from '../alt';
 
 class Actions {
-  updateListType(listType) {
```
EdwinGuzman: @ktp242 the list type is already part of the whole picks data set. Since there are a lot of pieces of information that gets updated, instead of updating each individually, it's better to just update the whole object once and get the data from the set object. So I removed this action and store value.

src/app/actions/BookActions.js
```diff
@@ -2,11 +2,7 @@
...
-    return listType;
-  }
-
-  updatePicks(picks) {
+  updatePicksData(picks) {
```
EdwinGuzman: Updated this to be the whole data set instead of just the array of picks.

src/app/components/Application/Main.jsx
```diff
@@ -1,33 +1,45 @@
...
 import utils from '../../utils/utils';
+import staffPicksDate from '../../utils/DateService';
+import appConfig from '../../../../appConfig';
 
 class Main extends React.Component {
   constructor(props) {
     super(props);
 
+    const picksData = this.props.picksData;
+    const picks =
```
EdwinGuzman: The picks always need to be filtered when they are set initially by the age type. This will not be the case for `childrens` or `ya` lists the function covers that.

src/app/components/Application/Main.jsx
```diff
@@ -1,33 +1,45 @@
...
     this.setSelectedFilter = this.setSelectedFilter.bind(this);
     this.clearFilters = this.clearFilters.bind(this);
+    this.getPicksInfo = this.getPicksInfo.bind(this);
+    this.getNewPickSet = this.getNewPickSet.bind(this);
+    this.filterByAudience = this.filterByAudience.bind(this);
+    this.getCount = this.getCount.bind(this);
   }
 
-  componentWillReceiveProps() {
+  componentWillReceiveProps(nextProps) {
```
EdwinGuzman: @ktp242 I also noticed you set this up which is okay but it's not used properly, it should have some kind of logic check that returns true or false. If the `nextProps` don't match the current `props`, then you _want_ to update the component. Otherwise, if they're the same you do not need to update and re-render the component. We can update this later to conform to the lifecycle rule.

src/app/components/Application/Main.jsx
```diff
@@ -39,7 +51,13 @@ class Main extends React.Component {
    */
   getNewPickSet(picks, selectedFilters) {
     if (!selectedFilters || !selectedFilters.length) {
-      return picks || [];
+      const updatedPicks =
+        this.filterByAudience(picks, this.props.currentAudience, 'staff-picks');
```
EdwinGuzman: Whenever we get a new pick set, we always want to filter it by audience type. The `'staff-picks'` should be updated later.

src/app/components/Application/Main.jsx
```diff
@@ -48,15 +66,30 @@ class Main extends React.Component {
       // Get the array of selected tags found in the book item
       const inSelectedFilter = utils.getSelectedTags(tagArray, selectedFilters);
 
-      if (inSelectedFilter.length &&
-          (inSelectedFilter.length === selectedFilters.length)) {
-        return book;
-      }
-
-      return undefined;
+      return (inSelectedFilter.length && (inSelectedFilter.length === selectedFilters.length));
```
EdwinGuzman: The filter function is better used if you return a boolean. If the boolean is true, the data object gets passed, otherwise it doesn't.

src/app/components/Application/Main.jsx
```diff
@@ -74,67 +107,85 @@ class Main extends React.Component {
...
+    if (document.getElementById('list-title')) {
+      setTimeout(() => {
+        document.getElementById('list-title').focus();
+      }, 400);
+    }
   }
 
   /**
-   * extractAudienceGroup(picks, audience, listType)
+   * filterByAudience(picks, audience, type)
```
EdwinGuzman: Updated the function name because it's more of a filter than extracting objects.

src/app/components/BookList/BookList.jsx
```diff
@@ -4,16 +4,28 @@ import PropTypes from 'prop-types';
...
       : null
   );
 
+  const { displayDate = {}, displayAge } = displayInfo;
+  const booksfound = `${picksCount} Book${picksCount === 1 ? '' : 's'} Found`;
+
   return (
     <div className="booklist-section nypl-column-three-quarters">
-      <h2>2017 Picks</h2>
+      <h2 id="list-title" tabIndex="0">
```
EdwinGuzman: @ktp242 I'll extract this out into another component so you can use it for the sidebar.

src/app/components/ListSelector/ListSelector.jsx
```diff
@@ -67,11 +66,17 @@ class ListSelector extends React.Component {
           // Leads the user to the 404 page
           this.updateLocation(`${config.baseUrl}404`);
         } else {
+          const data = response.data;
+          const filters = utils.getAllTags(data.picksData.picks);
```
EdwinGuzman: Once we get the new data from the server on a client call, we want to update the filters.

src/app/stores/BookStore.js
```diff
@@ -16,9 +15,8 @@ class BookStore {
     });
 
     this.state = {
-      listType: '',
       filters: [],
-      currentPicks: {},
+      picksData: {},
```
EdwinGuzman: This is a big change.

EdwinGuzman:
There is one issue with this PR which I intentionally left out. When you are on an age group like `Adult` and you switch to `Teen` or `Children`, the picks and title updates, but the filters do not. I was hoping someone else can pick that up to learn about the filtering done.

src/app/components/BookList/BookList.jsx
ktp242: Thanks!

PR #185, opened 2018-05-04
Add fade in anmiation to submit button to delay visibility

ktp242:
Can be tested here
http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-movies/recommendations/best-books/staff-picks/2018-01-01/

related ticket
https://jira.nypl.org/browse/WWW-242

This PR implements the animation fade-in on the styles, so it can delay the button's visibility until `componentDidMount` kicks in.

The problem was because `componentDidMount` will be executed only after all the DOMs are rendered. And only in `componentDidMount`, we set the value `isJsEnabled` to be true in the store. So we saw the submit button was rendered while `IsJsEnabled` was still false.


ktp242:
I did not mean using it in `componentDidMount` is a problem.

Maybe the way I delivered it makes the miscommunication. What I mean the issue is caused because `componentDidMoun`t will be executed only after all the DOMs are rendered. And that is merely a fact.

ktp242:
Update the the animation settings.

Tried to make a file to import it but it seemed to take longer time to load the styles and it did not apply the animation in time. It will take longer seconds for the fade in so ended up having it in the main.scss

EdwinGuzman:
That's interesting. I just noticed that `@nypl/dgx-header-component` has a `fadeIn` keyframe animation defined. So maybe it would fight that animation style. I also noticed that you copied and pasted the animation code AFTER the `sidebar.scss` and `booklist.scss` which use the animations. I'm not sure why they're still working if the keyframe code is defined AFTER.

Did you try to add it to a new file and import it before the SCSS for component files?

ktp242:
yeah that's exactly what I did. I made `animations.scss` and imported it in as `color.scss`, however, it seemed to delay the time it read the file, so it needed quite a long time for fade in to achieve the result we want.

ktp242:
I can try name it a different way so it won't be against the styles from header.

ktp242:
Updated! Thanks for the insight! I found that with two `fadeIn` from header and the application itself messed up the effect. I think they added up the time. I have changed the name for the local fade in animation and imported in from a independent file. It works well now.

EdwinGuzman:
Great!

PR #186, opened 2018-05-07
Www 214 dynamic filters

holingpoon:
This PR should fix the problem of Filter by Tags not changing when a user switches `audience` within a `season` for `staff-picks`. Please review. Thanks!

### DIFF
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -17,7 +17,6 @@ class Main extends React.Component {
       this.filterByAudience(picksData.picks, this.props.currentAudience, picksData.type);
 
     this.state = {
-      selectableFilters: this.props.selectableFilters,
       selectedFilters: [],
       picks,
     };
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -22,6 +22,7 @@ class Main extends React.Component {
       picks,
     };
 
+    this.setSelectableFilter = this.setSelectableFilter.bind(this);
     this.setSelectedFilter = this.setSelectedFilter.bind(this);
     this.clearFilters = this.clearFilters.bind(this);
     this.getPicksInfo = this.getPicksInfo.bind(this);
@@ -119,6 +120,9 @@ class Main extends React.Component {
     });
   }
 
+  setSelectableFilter(filterId, active) {
+  }
+
   /**
    * getCount(picksData = {})
    * Returns the count of the current set of picks.
@@ -179,6 +183,8 @@ class Main extends React.Component {
       });
     }
 
+    this.setSelectableFilter();
+
     return updatedPicks;
   }
 
@@ -190,6 +196,7 @@ class Main extends React.Component {
       <div className="nypl-row">
         <Sidebar
           filters={this.props.filters}
+          // TODO: Selectable Filters needs to be refreshed
           selectableFilters={this.state.selectableFilters}
           setSelectedFilter={this.setSelectedFilter}
           clearFilters={this.clearFilters}
@@ -215,6 +222,7 @@ class Main extends React.Component {
 Main.propTypes = {
   filters: PropTypes.array,
   selectableFilters: PropTypes.array,
+  setSelectableFilter: PropTypes.func,
   picksData: PropTypes.object,
   isJsEnabled: PropTypes.bool,
   listOptions: PropTypes.object,
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -22,7 +22,7 @@ class Main extends React.Component {
       picks,
     };
 
-    this.setSelectableFilter = this.setSelectableFilter.bind(this);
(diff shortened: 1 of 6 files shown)


src/app/utils/utils.js
```diff
@@ -66,6 +66,20 @@ function Utils() {
     return selectedTags;
   };
 
+  /**
+   * getFilters(filters)
+   * Generates a Map of (pick tag id => pick tag label) key-value pairs,
+   * returns an Object of this map.
+   * @param filters
```
EdwinGuzman: Update this to follow https://esdoc.org/ convention.

src/app/utils/utils.js
```diff
@@ -66,6 +66,20 @@ function Utils() {
     return selectedTags;
   };
 
+  /**
+   * getFilters(filters)
```
EdwinGuzman: You're not really getting filters, you're mapping the filter array to id/label pairs, so maybe a better name?

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -37,7 +35,8 @@ class BookFilters extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({ selectedFilters: nextProps.selectedFilters });
+    const newFilters = utils.getFilters(nextProps.filters);
```
EdwinGuzman: You can rename this variable to be `filters`. For creating objects, you can use the shortcut:
```
{
  selectedFilters: nextProps.selectedFilters,
  filters,
}
```

In the example above, the key is `filters` but since there's a variable with the same name, you do not have to do `filters: filters`.

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -166,6 +167,7 @@ BookFilters.propTypes = {
   filters: PropTypes.array,
   selectableFilters: PropTypes.array,
   setSelectedFilter: PropTypes.func,
+  setSelectableFilters: PropTypes.func,
```
EdwinGuzman: Why did you add this here?

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -17,15 +17,13 @@ class BookFilters extends React.Component {
...
-        id: filter.toLowerCase().split(' ').join('-'),
-        label: filter,
-      })),
+      filters: utils.getFilters(this.props.filters),
       selectedFilters: this.props.selectedFilters,
       focusId: '',
       disabled: false,
       showFilters: false,
       timeout: undefined,
+      selectableFilters: this.props.selectableFilters,
```
EdwinGuzman: Why did you add this here?

src/app/components/Application/Main.jsx
```diff
@@ -182,15 +182,15 @@ class Main extends React.Component {
...
+
   render() {
     const { type } = this.props.picksData;
     const picksCount = this.getCount();
-
     return (
       <div className="nypl-row">
         <Sidebar
           filters={this.props.filters}
-          selectableFilters={this.state.selectableFilters}
```
EdwinGuzman: Since you're not using this anymore, you can remove this from the component.

src/app/components/BookFilters/BookFilters.jsx
ktp242: or you can just do `filters: utils.getFilters(nextProps.filters)` if you don't need to assign `utils.getFilters(nextProps.filters)` anywhere else.

src/app/components/Application/Main.jsx
holingpoon: Removed `this.state.selectableFilters` from `Main`'s constructor.

EdwinGuzman: Right, it's not just there... Check the file again to see other instances of it in the state that are not being used.

holingpoon:
I have merged `development` into this PR and resolved the merge conflicts.

src/app/components/BookFilters/BookFilters.jsx
```diff
@@ -37,7 +34,10 @@ class BookFilters extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({ selectedFilters: nextProps.selectedFilters });
+    this.setState({
+      selectedFilters: nextProps.selectedFilters,
+      filters: utils.getFiltersMapping(nextProps.filters)
```
ktp242: Missing a trailing comma here

test/unit/utils.test.js
```diff
@@ -163,6 +183,7 @@ describe('Utils functions', () => {
       expect(mockElements[1].focus.callCount).to.equal(0);
       expect(mockElements[2].focus.callCount).to.equal(1);
       expect(mockElements[3].focus.callCount).to.equal(0);
+
```
ktp242: I don't think we need an additional line here?

ktp242:
Small syntax stuff! But it looks good!

PR #188, opened 2018-05-09
Annual lists updates

gkallenberg:
Adds routing, component, config and unit test changes for annual list rendering with the new selector component. There are probably unit test missing, so feel free to suggest what else needs coverage.

### DIFF
--- a/appConfig.js
+++ b/appConfig.js
@@ -2,6 +2,7 @@ const config = {
   port: 3001,
   webpackDevServerPort: 3000,
   baseUrl: '/books-music-movies/recommendations/best-books/',
+  base404: '/books-music-movies/recommendations/best-books/404',
   baseApiUrl: '/books-music-movies/recommendations/best-books/api/',
   metaTags: {
     ya: [
@@ -126,6 +127,13 @@ const config = {
         { name: 'Teen', value: 'YA' },
         { name: 'Children', value: 'Children' },
       ],
+    },
+  },
+  annualListOptions: {
+    season: {
+      fieldsetName: 'season',
+      currentValue: '',
+      options: [],
     },
   },
   audienceMap: {
--- a/src/app/components/About/About.jsx
+++ b/src/app/components/About/About.jsx
@@ -38,8 +38,8 @@ class About extends React.Component {
     return (
       <div className="about nypl-row">
         <h2>Additional Information</h2>
-        <p>Many of these titles are available in formats for <a href={aboutUrls.print}>patrons
-        with print disabilities.</a>
+        <p>Many of these titles are available in formats for
+          <a href={aboutUrls.print}>patrons with print disabilities.</a>
         </p>
         <p><a href={aboutUrls.about}>About Best Books</a></p>
         <ul>
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -82,12 +82,10 @@ class Main extends React.Component {
     const { date } = picksData;
     const { type } = picksData;
     const displayDate = type === 'staff-picks' ? staffPicksDate(date) : annualDate(date);
-    const displayAudience = !['teens', 'kids'].includes(type) ? 'Adult' : utils.capitalize(type);
 
     return {
       displayDate,
-      displayAge: (appConfig.audienceMap[currentAudience]) ?
-        appConfig.audienceMap[currentAudience] : displayAudience,
+      displayAge: appConfig.audienceMap[currentAudience || 'Adult'],
     };
   }
 
@@ -178,7 +176,6 @@ class Main extends React.Component {
   }
 
   render() {
-    const { type } = this.props.picksData;
     const picksCount = this.getCount();
     return (
       <div className="nypl-row">
@@ -200,7 +197,7 @@ class Main extends React.Component {
         <BookList
           picks={this.state.picks}
           isJsEnabled={this.props.isJsEnabled}
-          displayType={type}
+          displayType={this.props.picksData.type}
           displayInfo={this.getPicksInfo(this.props.picksData, this.props.currentAudience)}
           picksCount={picksCount}
         />
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -86,7 +86,8 @@ class Main extends React.Component {
 
     return {
       displayDate,
-      displayAge: (appConfig.audienceMap[currentAudience]) ? appConfig.audienceMap[currentAudience] : displayAudience,
+      displayAge: (appConfig.audienceMap[currentAudience]) ?
+        appConfig.audienceMap[currentAudience] : displayAudience,
     };
   }
 
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import BookList from '../BookList/BookList';
 import Sidebar from '../Sidebar/Sidebar';
 import utils from '../../utils/utils';
-import staffPicksDate from '../../utils/DateService';
+import { staffPicksDate, annualDate } from '../../utils/DateService';
 import appConfig from '../../../../appConfig';
 
 class Main extends React.Component {
@@ -82,11 +82,13 @@ class Main extends React.Component {
     }
 
     const { date } = picksData;
-    const displayDate = staffPicksDate(date);
+    const { type } = picksData;
+    const displayDate = type === 'staff-picks' ? staffPicksDate(date) : annualDate(date);
+    const displayAudience = !['teens', 'kids'].includes(type) ? 'Adult' : utils.capitalize(type);
 
     return {
       displayDate,
-      displayAge: appConfig.audienceMap[currentAudience || 'Adult'],
+      displayAge: (appConfig.audienceMap[currentAudience]) ? appConfig.audienceMap[currentAudience] : displayAudience,
     };
   }
 
@@ -198,6 +200,7 @@ class Main extends React.Component {
           currentAudience={this.props.currentAudience}
           displayInfo={this.getPicksInfo(this.props.picksData, this.props.currentAudience)}
           picksCount={picksCount}
+          type={this.props.picksData.type}
         />
 
         <BookList
--- a/src/app/components/Application/Main.jsx
+++ b/src/app/components/Application/Main.jsx
@@ -176,7 +176,6 @@ class Main extends React.Component {
     return updatedPicks;
   }
 
-
   render() {
     const { type } = this.props.picksData;
     const picksCount = this.getCount();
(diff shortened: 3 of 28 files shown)


src/app/utils/DateService.js
```diff
@@ -23,21 +24,39 @@ function monthOrSeason(month, year) {
   return (year >= 2016) ? season : monthsArr[month];
 }
 
-/** staffPicksDate(date)
```
gkallenberg: Add to following line.

src/server/routes/annualData.js
```diff
@@ -1,62 +1,192 @@
...
+
+import utils from '../../app/utils/utils';
+import platformConfig from '../../../platformConfig';
+import modelListOptions from '../../app/utils/ModelListOptionsService';
+import { matchListDate } from '../../app/utils/DateService';
 
 const nyplApiClientGet = endpoint =>
   nyplApiClient().then(client => client.get(endpoint, { cache: false }));
 
-/* annualCurrentData
```
gkallenberg: Add to following line after comment start

src/server/routes/annualData.js
```diff
@@ -1,62 +1,192 @@
...
     .catch((error) => {
-      console.log(`Error fetching endpoint: ${error}`);
+      console.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+
+      return res.redirect(`${config.baseUrl}404`);
+    });
+}
+
+/**
+ * Get a specific month's or season's staff pick list.
```
gkallenberg: Add function name

src/app/utils/DateService.js
```diff
@@ -1,6 +1,7 @@
 import {
   findKey as _findKey,
   contains as _contains,
+  isNaN as _isNaN,
```
gkallenberg: Remove since not used.

holingpoon:
@gkallenberg Pulled branch and try to run on my local instance. The title is taking over `staff-picks`, so only "2018 Picks" instead of "Spring 2018 Picks for Adult", and Spring 2018 Young Adults staff picks return no records.

src/app/components/Application/Main.jsx
```diff
@@ -192,6 +194,7 @@ class Main extends React.Component {
           currentAudience={this.props.currentAudience}
           displayInfo={this.getPicksInfo(this.props.picksData, this.props.currentAudience)}
           picksCount={picksCount}
+          type={this.props.picksData.type}
```
gkallenberg: Change to displayType. Fixes issue with Staff Picks titles.

gkallenberg:
@holingpoon -- looks like I didn't name the prop correctly in Main.jsx. Will fix ASAP.

appConfig.js
```diff
@@ -2,6 +2,7 @@ const config = {
   port: 3001,
   webpackDevServerPort: 3000,
   baseUrl: '/books-music-movies/recommendations/best-books/',
+  base404: '/books-music-movies/recommendations/best-books/404',
```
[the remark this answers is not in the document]
ktp242: I think it is OK. But if we are adding it to the config, then we should just use it from config instead of typing it whenredirecting 404 page. I haven't check all the files yet so please ignore me if you have already done it in this PR :-P

src/app/components/Application/Main.jsx
```diff
@@ -80,11 +80,13 @@ class Main extends React.Component {
...
     const { date } = picksData;
-    const displayDate = staffPicksDate(date);
+    const { type } = picksData;
+    const displayDate = type === 'staff-picks' ? staffPicksDate(date) : annualDate(date);
+    const displayAudience = !['teens', 'kids'].includes(type) ? 'Adult' : utils.capitalize(type);
 
     return {
       displayDate,
-      displayAge: appConfig.audienceMap[currentAudience || 'Adult'],
+      displayAge: (appConfig.audienceMap[currentAudience]) ? appConfig.audienceMap[currentAudience] : displayAudience,
```
ktp242: Can we make this line shorter so eslint will be happy?

src/app/components/BookList/BookList.jsx
```diff
@@ -10,18 +10,19 @@ const BookList = ({
   picks,
   type,
   displayInfo,
+  displayType,
   picksCount,
 }) => {
   const renderBookItems = currentBooks => (
     currentBooks.length ?
-      currentBooks.map(book => <Book key={book.slug} pick={book} isJsEnabled={isJsEnabled} />)
```
ktp242: If we delete `key` here it will still work, but we will receive the complaints from React on bowser console.

src/app/components/ListSelector/ListSelector.jsx
```diff
@@ -55,9 +55,11 @@ class ListSelector extends React.Component {
       console.log('No valid season input.');
       return;
     }
+    const type = this.props.fieldsetProps.type;
 
+    console.log(`${config.baseApiUrl}${type}/${submitValue}`);
```
ktp242: Do we need this console?

src/app/utils/DateService.js
ktp242: I thought you added it not removed it?

src/app/components/Sidebar/Sidebar.jsx
```diff
@@ -85,6 +90,7 @@ Sidebar.propTypes = {
   listOptions: PropTypes.object,
   selectedFilters: PropTypes.arrayOf(PropTypes.object),
   displayInfo: PropTypes.object,
+  type: PropTypes.string,
```
ktp242: We miss the type validation for `currentType`

gkallenberg: Changed to type -- currentType is no longer the prop name.

gkallenberg:
@ktp242 and @holingpoon -- fixed most of the things and have the no-js version working properly, I believe.

ktp242:
Works nicely!!

src/app/components/About/About.jsx
```diff
@@ -38,8 +38,8 @@ class About extends React.Component {
     return (
       <div className="about nypl-row">
         <h2>Additional Information</h2>
-        <p>Many of these titles are available in formats for <a href={aboutUrls.print}>patrons
-        with print disabilities.</a>
+        <p>Many of these titles are available in formats for
```
EdwinGuzman: No need to worry about this. I believe it will be removed.

ktp242:
at `http://local.nypl.org:3001/books-music-movies/recommendations/best-books/childrens/2014`
I ran into this warning: `flattenChildren(...): Encountered two children with the same key, `1:$Families`. Child keys must be unique; when two children share a key, only the first child will be used`

I think this comes from filter button 's DOM. I believe there's a second key as Families.

holingpoon:
For me, it occurs when switching into `ya/2015` from any year, and `Adventures` seems to have duplicate entries when `flattenChildren()` was running.

src/app/components/Application/Main.jsx
EdwinGuzman: What does your comment mean? Also, on line 179 `type` is defined in `const { type } = this.props.picksData;`. That's the same (but different syntax) for `const type = this.props.picksData.type;`

src/app/components/Book/Book.jsx
```diff
@@ -87,13 +87,17 @@ const Book = ({ pick, isJsEnabled }) => {
...
-
-      return (
+      let reviewerName = '';
+      let location = '';
+      if (review.reviewerName) {
+        reviewerName = review.reviewerName.trim();
+        location = review.reviewerLocation ? `, ${review.reviewerLocation.prefLabel}` : '';
+      }
+
+      return (!reviewerName) ? (<div><p className="book-item-description">{text}</p></div>) : (
```
EdwinGuzman: Interesting way of doing this. My preferred way of doing this would be something like 
```
const reviewerNameDOM = reviewerName ? (<p className="book-item-picked-by">Staff Pick By: {reviewerName}{location}</p>) : null;

return (
  <div>
    <p className="book-item-description">{text}</p>
    {reviewerNameDOM}
  </div>
);
```

just because you then don't have to write out the DOM for the `book-item-description` twice.

EdwinGuzman: And if an element is `null`, nothing will get rendered.

EdwinGuzman:
The unique key issue is related to assigning unique keys to elements in an array when they are rendered. The value to the `key={...}` is no longer unique and causing that warning.

src/app/components/Application/Main.jsx
gkallenberg: I'm confusing my variables here. Looks like I need to just pick a name and use that consistently. Was thinking of using `displayType` to use a similar construct to `displayInfo` but should just use type instead. I'll remove the extra assignment at line 179 -- leftover from testing variable outputs, I believe.

gkallenberg:
For the key issue, I swapped the iterator for key to tag which causes the duplicate issue. I did this because the linter complained about using iterator as key not being allowed. Perhaps a better combination here would be better? Tag plus iterator? Not sure what the best practice is for things like this in React.

ktp242:
I run into the warning when choosing a filter on `http://local.nypl.org:3001/books-music-movies/recommendations/best-books/childrens/2014`

`Warning: Failed prop type: Invalid prop `selectedFilters[0]` of type `string` supplied to `Sidebar`, expected `object`.`

gkallenberg:
@ktp242, I see that error in all other branches as well which predates this branch. Wondering if this is being addressed by @holingpoon's changes for filters??

ktp242:
OK! Let's isolate that issue for another PR since it is not from the updates here.

src/app/components/Application/Main.jsx
EdwinGuzman: Cool, leaving the variable there and just using `type={type}` would've been okay, too.

src/server/routes/annualData.js
```diff
@@ -1,62 +1,195 @@
...
+ * annualClientListData(req, res, next)
+ * Gets a specific month's or season's staff pick list on the client side.
+ * @param {object} req
+ * @param {object} res
+ */
+function annualClientListData(req, res) {
+  const dataType = utils.getDataType(req.params.type);
+  const seasonMatches = matchListDate(req.params.time, req.params.type);
+
+  if (!seasonMatches) {
```
EdwinGuzman: This should be `year` since `seasons` aren't used in annual lists.

src/server/routes/annualData.js
```diff
@@ -1,62 +1,195 @@
...
+      console.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+
+      return res.redirect(`${config.baseUrl}404`);
+    });
+}
+
+/**
+ * annualListData(req, res, next)
+ * Get a specific month's or season's staff pick list.
+ * It calls '/book-lists?type=staff-picks' to get all the available list options first.
```
EdwinGuzman: Some of the comments and variables should be updated to reflect the annual descriptions.

PR #189, opened 2018-05-09
Kp/www 222 remove about best books link

ktp242:
This PR adds the check for the list type, and based on the result to add or not add "About Best Books" link.

### DIFF
--- a/src/app/components/About/About.jsx
+++ b/src/app/components/About/About.jsx
@@ -36,7 +36,8 @@ class About extends React.Component {
     const shareUrl = this.state.location.href;
     const shareText = encodeURI(pageTitle[type]);
     const aboutBestBooksLink = ['teens', 'kids'].includes(type) ?
-      <p><a className="about-best-books-link" href={aboutUrls.about}>About Best Books</a></p> : null;
+      <p><a className="about-best-books-link" href={aboutUrls.about}>About Best Books</a></p>
+      : null;
 
     return (
       <div className="about nypl-row">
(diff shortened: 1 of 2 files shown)


ktp242:
The current one is `NYPL | Recommendations`. Is this the correct one to use?

EdwinGuzman:
Hmm no, never mind that. It seems my files weren't fully updated. Also, when I try the branch out locally, going to `.../best-books/ya` or `.../best-books/childrens` breaks but I'm not sure if it's because of this PR.

ktp242:
It should be fixed with Greg's PR 
https://github.com/NYPL/staff-picks/pull/188

appConfig.js
```diff
@@ -102,7 +102,7 @@ const config = {
   pageTitle: {
     ya: 'Best Books for Teens | NYPL',
     childrens: 'Best Books for Kids | NYPL',
-    'staff-picks': 'Staff Picks | NYPL',
+    'staff-picks': 'NYPL | Staff Picks',
```
gkallenberg: I don't think this change is necessary. Normally, we use the specific title and then pipe in NYPL, {title} | NYPL for all page titles.

EdwinGuzman: Hmmm. Yeah, also a question for Courtney but the content itself doesn't matter much for this PR.

EdwinGuzman:
Okay, I'm approving but I do think it should be merged until Greg's update is in `development`.

appConfig.js
ktp242: The title is from the answer of Courtney, so it should be correct. Since I have updated it then I will just keep it here.

ktp242:
Thanks for the reviewing. I can wait for Greg's PR merged then.

ktp242:
The PR has updated. It updated props name and related test to reflect the changes on development branch.

PR #190, opened 2018-05-10
Quick fix - making the image height bigger so that the byline doesn't…

EdwinGuzman:
… float down the card.

Quick fix for @ricardogalvez .

### DIFF
--- a/src/client/styles/components/_bookItem.scss
+++ b/src/client/styles/components/_bookItem.scss
@@ -40,11 +40,11 @@ $book-card-largescreen-gutter: 3.5%;
     float: left;
 
     @include min-screen($xtrasmall-breakpoint) {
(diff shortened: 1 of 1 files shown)


EdwinGuzman:
That I don't see. Screenshot?

gkallenberg:
Screen shots from the current Staff Picks Adult list.
screen shot 2018-05-10 at 12 58 13 pm
screen shot 2018-05-10 at 12 58 02 pm
screen shot 2018-05-10 at 12 57 51 pm

EdwinGuzman:
It's because the height is _still_ not big enough. I'll just make it bigger.

ktp242:
Codes look good. If @ricardoom feels OK with the outcome then I think it can be approved.

PR #191, opened 2018-05-10
Meta tags

EdwinGuzman:
This will get and render the correct page title and meta tags (OG and Twitter) on the server side. They are static and do not need to change between year or season.

This PR does NOT include any updates to images or descriptions, but I did copy over the current Staff Picks image and description from the current Staff Picks on Prod (nypl.org/staff-picks).

@courtneymcgee that can be a separate ticket unless if what's here already is fine. This will render the meta tags like so (and note, it's not related to the banner image):
screen shot 2018-05-10 at 1 19 33 pm

### DIFF
--- a/server.js
+++ b/server.js
@@ -81,8 +81,6 @@ app.use('/', (req, res) => {
 
       iso.add(html, alt.flush());
 
-      console.log(pageTitle, metaTags);
-
       res
         .status(200)
         .render('index', {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -76,11 +76,13 @@ app.use('/', (req, res) => {
         res.locals.data.metaTags : [];
       const renderedMetaTags = metaTags.map((tag, index) =>
         ReactDOMServer.renderToString(<meta key={index} {...tag} />));
-      const renderedPageTitle = (res.locals.data && res.locals.data.pageTitle) ?
+      const pageTitle = (res.locals.data && res.locals.data.pageTitle) ?
         res.locals.data.pageTitle : '';
 
       iso.add(html, alt.flush());
 
+      console.log(pageTitle, metaTags);
```
ktp242: remove the log?

EdwinGuzman: Yep.

PR #193, opened 2018-05-10
Kp/www 215 revise placeholder hero

ktp242:
This PR updates hero elements so now it has different heros for staff picks list and non-staff picks lists.

Ticket 215: https://jira.nypl.org/browse/WWW-215

The heros for non staff picks lists remain the same. However, the heros for staff picks,
**When on desktop view**,
1) has no background image but a solid red background color
2) the color of text will be white
3) has a section titile, `BOOKS/MUSIC/MOVIES`

screencapture-local-nypl-org-3001-books-music-movies-recommendations-best-books-2018-05-10-17_51_21

**When on small view**,
1) has no background image but a solid white background(non-background color)
2) the color of text will be black
3) dos not have a section titile, `BOOKS/MUSIC/MOVIES`

screencapture-local-nypl-org-3001-books-music-movies-recommendations-best-books-2018-05-10-17_50_06


ktp242:
It is on dev too! @ricardoom 
http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-movies/recommendations/best-books/staff-picks/2017-04-01

src/app/components/Hero/Hero.jsx
```diff
@@ -3,16 +3,26 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const Hero = ({ heroData }) => {
+  // Check if this list is one of "Staff Picks"
+  const isStaffPicksList = heroData.header === 'Staff Picks';
```
gkallenberg: Perhaps it would be better to use picksData.type as the test for determining if staff picks is being rendered rather than a hardcoded string.

src/app/components/Hero/Hero.jsx
```diff
@@ -3,16 +3,26 @@ import React from 'react';
...
   const image = heroData.heroImageUrl;
+  // "Staff Picks" lists have section title
+  const sectionTitle = isStaffPicksList ?
+    <p className="hero-section-title">{heroData.sectionTitle}</p> : null;
+  const staffPicksHeroClassName = isStaffPicksList ? ' staff-picks-hero' : '';
+  // "Staff Picks" lists do not have a hero image but a full red background color
+  const heroImageDOM = isStaffPicksList ? null :
+    <div className="hero-image">
+      <img src={image} alt="" />
+    </div>;
```
gkallenberg: Would it be better to define the `sectionTitle`, `staffPicksHeroClassName` and `heroImageDom` using `let` and do one conditional test for isStaffPicksList and set the variables accordingly?

Rename `sectionTitle` to `sectionTitleDOM`

test/unit/Hero.test.js
```diff
@@ -13,8 +14,105 @@ describe('Hero', () => {
...
+    });
+
+    it('should has the correct style for backgeround and texts.', () => {
+      expect(component.find('div.hero').node.props.className).to.equal('hero staff-picks-hero');
+    });
+
+    it('should not render the image.', () => {
+      expect(component.find('img').length).to.equal(0);
+    });
+  });
```
gkallenberg: Change `should has` to `should have` on these tests.

src/app/components/Hero/Hero.jsx
ktp242: sure!

src/app/components/Hero/Hero.jsx
EdwinGuzman: Agreed.

test/unit/Hero.test.js
ktp242: oh my! thanks for finding these.

src/client/styles/components/_hero.scss
```diff
@@ -64,4 +64,31 @@
...
+
+    .hero-content-description {
+      background-color: transparent;
+    }
+
+    @include media($mobile-breakpoint) {
+      background-color: #fff;
+      color: #000;
+
+      .hero-section-title {
```
EdwinGuzman: Isn't this the same as the visually hidden class rules? Why not use that class?

ktp242: sure!

src/app/components/Hero/Hero.jsx
gkallenberg: Just seems more consistent to me to pass type down to the Hero component and test type === 'staff-picks' instead of the header's title.

ktp242:
Updated the logic to check if the list belongs to staff picks list up to Application.jsx

PR #195, opened 2018-05-11
Www 258 logging winston

holingpoon:
This PR should set up Winston module to log server-side messages. It is based on the setup in `NYPL-discovery/discovery-front-end` repository.


holingpoon:
@kfriedman Like this one? https://github.com/NYPL/dgx-loggly

kfriedman:
Yes, kind of. But it could be much lighter weight. 

Just basically including the functionality from `logger.js`.

EdwinGuzman:
Making the logger a module is a good idea. I was thinking the same when I told @holingpoon that it's very similar to what's in Discovery now. It's similar to `dgx-loggly` except that it would not go to loggly, it'd be a specific node-base logger to Winston.

logger.js
```diff
@@ -0,0 +1,92 @@
...
+    case 'info':
+      return 6;
+    case 'debug':
+      return 7;
+    default:
+      return 'n/a';
+  }
+};
+
+// const logLevel = (process.env.NODE_ENV === 'production') ? 'info' : 'debug';
```
EdwinGuzman: It looks like we don't need this so it can be removed.

ktp242:
Looking good.

EdwinGuzman:
The tests are fine. They could be more in-depth about what params get passed and formats for the JSON response, but maybe later when it becomes its own module.

PR #197, opened 2018-05-12
Add react lazy

ktp242:
This PR adds lazy load for loading the images while rendering the web page. The related ticket is https://jira.nypl.org/browse/WWW-257

Lazy load improves the performance a lot. And to reduce the trade for image flickering, the PR adds cushion, in the range of height the images will be loaded even they are not visible yet, and fade in transition when the images are loaded.


src/client/styles/components/_bookItem.scss
```diff
@@ -38,6 +38,7 @@ $book-card-largescreen-gutter: 3.5%;
     display: block;
     max-width: 30%;
     float: left;
+    width: 30%;
```
ktp242: Set the fix width as placeholder so when the image is loaded the layout won't shift.

test/unit/Book.test.js
```diff
@@ -41,18 +41,18 @@ describe('Book Component', () => {
...
     it('should not render the Book component if the pick object prop is empty', () => {
-      component = mount(<Book pick={{}} />);
+      component = shallow(<Book pick={{}} />);
     });
   });
 
   describe('Component with data and JavaScript enabled', () => {
     before(() => {
-      component = mount(<Book pick={pickObject} isJsEnabled />);
+      component = shallow(<Book pick={pickObject} isJsEnabled />);
```
ktp242: Use `shallow` instead of `mount` so the test won't be affected by <Lazy>.

PR #198, opened 2018-05-14
Update changelog

ktp242:
This PR updates the `CHANGELOG.md` because we made bunch of updates. Plz let me know if I missed anything. Or help me change better words lol

This PR will be merged after all our current tickets for this sprint are merged.


CHANGELOG.md
```diff
@@ -2,9 +2,12 @@
...
-> Update the BookPage component to support legacy staff-picks direct links and homepage's staff-picks component as the modal replacement from the previous staff-picks application. No direct links to picks in lists is supported at this time.
+> Add an ID on each book item and update the functions for handling their routes. If a patron adds the hash that consists of a book's name and its ISBN to the URL, it will scroll down and focus on the book. Same as the routes will do with the book's name and ISBN as a path. The pages for individual books are no longer supported.
 > Add an `.env.example` and update the related README.md so now the developer can set a `.env` to run the application locally.
 > Add a list selector component that consists of two dropdown menus for choosing both seasons and audience from Staff Picks, or only seasons from Best YA Books and Best Children Books.
+> Update hero elements and styles based on different list types. Staff Picks and Children/YA lists now have different heros.
+> Add lazy load when loading images for better performance.
+> Add winston and logging.
```
EdwinGuzman: Capitalize Winston.

CHANGELOG.md
```diff
@@ -2,9 +2,12 @@
 
 ### 4.0.0
 > Update the routes. The base route is now `/books-music-movies/recommendations/best-books/`. And `staff-picks`, `ya`, `childrens` are three lists on the same level of the structure.
-> Update the BookPage component to support legacy staff-picks direct links and homepage's staff-picks component as the modal replacement from the previous staff-picks application. No direct links to picks in lists is supported at this time.
+> Add an ID on each book item and update the functions for handling their routes. If a patron adds the hash that consists of a book's name and its ISBN to the URL, it will scroll down and focus on the book. Same as the routes will do with the book's name and ISBN as a path. The pages for individual books are no longer supported.
```
EdwinGuzman: Hmm not `Add`, more like `Update` the function routes that handle book item ids. As we've said before, we are not linking to individual books and patrons will not be able to create URLs. So they won't know that they need to add an ISBN and its title to create the URL. Instead, the description should say that the hash URL is for supporting old links that currently exist. Also, you can add that old links with the id will redirect to the hashed url.

Thanks for writing this. I was going to write this up later on before making the release.

ktp242: Sure! OK, how about this?

`Update the routes for supporting the old URLs for individual books. Now the links no longer lead to a modal but scroll down and focus on to the particular book on the list.`

EdwinGuzman: Good, just remove `to` and update `in` instead of `on` in "... focus on ~~to~~ the particular book ~~on~~in the list".

ktp242:
Updated the new description based on the comments for new routes of individual books.

PR #199, opened 2018-05-14
Fix 404 routing

gkallenberg:
Update routes.jsx to use named routes instead of `:type`
Use location.pathname in Application.jsx instead of `:type` which is no longer defined in routes.jsx
Update selectData.js to pass `next()` instead of redirecting to the baseUrl or base404 endpoints
Add logic to remove the Hero for 404 pages


src/app/routes/routes.jsx
```diff
@@ -1,25 +1,32 @@
...
-      <Route path={`${appConfig.baseUrl}404`} component={Error404Page} />
-      <Route path={`${appConfig.baseUrl}:type`} component={Main} />
-      <Route path={`${appConfig.baseUrl}:type/:time`} component={Main} />
-      <Redirect from="*" to={`${appConfig.baseUrl}404`} />
-    </Route>
+    <Router>
+      <Route path={`${config.baseUrl}`} component={App}>
+        <IndexRoute component={Main} />
+        <Route path={`${config.base404}`} status={404} component={Error404Page} />
+        <Route path={`${config.baseUrl}staff-picks`} component={Main} />
```
EdwinGuzman: This doesn't work if you remove lines 21-23 and make the `:time` param optional?

src/app/components/Application/Application.jsx
```diff
@@ -38,13 +38,19 @@ class App extends React.Component {
 
   render() {
     let heroData;
-    const type = !_isEmpty(this.props.params) ? this.props.params.type : undefined;
+    let heroDOM;
+    // Use pathname index as type since routes.jsx specifies them directly.
+    let type = this.props.location.pathname.split('/')[4];
```
ktp242: Eslint complains 'location' is missing in props validation.

src/app/routes/routes.jsx
```diff
@@ -1,25 +1,32 @@
 // React library
 import React from 'react';
 // Import Router
-import { IndexRoute, Route, Redirect } from 'react-router';
+import { IndexRoute, Router, Route, Redirect } from 'react-router';
```
ktp242: `Redirect` is never used. We can remove it.

src/server/routes/selectData.js
```diff
@@ -1,10 +1,7 @@
 import monthData from './monthData';
 import annualData from './annualData';
-import appConfig from '../../../appConfig';
```
ktp242: I think we will need `appConfig`. It is used at Line 18 and 21.

gkallenberg: Yup -- will add back -- hasty merge.

src/app/components/Application/Application.jsx
ktp242: Probably to have a check here to see if `this.props.location.pathname` is a string.

src/app/components/Application/Application.jsx
```diff
@@ -38,13 +38,19 @@ class App extends React.Component {
...
+      type = 'staffPicks';
+    }
 
     // Check if the params type is included in valid data set
     if (type && _allKeys(config.heroData).includes(type)) {
-      heroData = config.heroData[this.props.params.type];
-    } else {
-      heroData = config.heroData.staffPicks;
+      heroData = config.heroData[type] ? config.heroData[type] : config.heroData.staffPicks;
+      heroDOM = <Hero heroData={heroData} />;
```
ktp242: As the development branch updated. This part of scripts will need be updated too. For example, <Hero> now takes two props.

ktp242:
Some situations I noticed

1. if we have the URL like this
http://local.nypl.org:3001/books-music-movies/recommendations/best-books/staff-picks/2017-06-01/fjlnfhlf%20jhnh
it will redirect to
http://local.nypl.org:3001/books-music-movies/recommendations/best-books/staff-picks/2017-06-01#fjlnfhlf%20jhnh
and still render the list
2. if we have the URL like this
http://local.nypl.org:3001/books-music-movies/recommendations/best-books/ya/2014/igjflglfjlgfgjlfg
it will redirect to http://local.nypl.org:3001/books-music-movies/recommendations/best-books/ya/2014#igjflglfjlgfgjlfg
and break the application
3. if we have URL like this
http://local.nypl.org:3001/books-music-movies/recommendations/best-books/hdfkjghkfghkfhg
this will lead to 404 content but the URL still remains

src/server/routes/selectData.js
gkallenberg: Added back. See below.

EdwinGuzman:
To follow up on Kang's last comment, you can fix that by adding a check in `componentDidMount` in `main.jsx` to make sure that `pick` exists in line 39/40. That way, if pick doesn't exist in the list, then redirect because the user probably mistyped the id.

gkallenberg:
I've addressed issues #1 and #2 of @ktp242's comment. #3 I think is fine the way it is for now. It may need a more practiced hand to add the redirect to the 404 somewhere where it doesn't re-introduce the 404 redirect loop. Pages on the main site typically serve the 404 content with the URL remaining as-is. See: https://www.nypl.org/lakjlskje

server.js
```diff
@@ -54,6 +54,16 @@ nyplApiClient();
 
 app.use('/', expressRoutes);
 
+// Handle trailing slash
+app.use((req, res, next) => {
+  const test = /\?[^]*\//.test(req.url);
```
EdwinGuzman: A better variable name would be more helpful.

PR #206, opened 2018-05-16
Year month and selector changes

gkallenberg:
These help to resolve issues #200 and #201. #200 is a simple lable change for the annual lists for Sidebar dropdowns. Should display 'Year' instead of 'Season.'

Changes for issue #201 requires coordination in updating the metadata for lists on the ingester side. Once the year-month pattern for lists is available in the API, the direct links and redirects should work. 

To test, start your local instance using APP_ENV=production. The new pattern is deployed there for staff-picks lists. There are 2 lists there that you can use to test the code changes within this PR.

### DIFF
--- a/src/server/routes/monthData.js
+++ b/src/server/routes/monthData.js
@@ -68,6 +68,18 @@ function currentMonthData(req, res, next) {
     });
 }
 
+/**
+ * Returns first two matches hyphenated for redirecting a three part date
+ * to a two part date in a URL.
+ * @param {object} match
+ * @param {integer} d1
+ * @param {integer} d2
+ * @return {string}
+ */
+function toMonthAndYear(match, d1, d2) {
+  return [d1, d2].join('-');
+}
+
 /* selectMonthData
  * Get a specific month's or season's staff pick list.
 * It calls '/book-lists?type=staff-picks' to get all the available list options first.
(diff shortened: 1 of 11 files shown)


ktp242:
Let's update the data model for`test/unit/ListFilter.test.js` and `test/unit/ListSelector.test.js` too.

src/server/routes/monthData.js
```diff
@@ -68,16 +68,35 @@ function currentMonthData(req, res, next) {
...
 
+/**
+ * Returns first two matches hyphenated for redirecting a three part date
+ * to a two part date in a URL.
+ * @param {object} match
+ * @param {integer} d1
+ * @param {integer} d2
+ * @return {string}
+ */
+function toMonthAndYear(match, d1, d2) {
```
ktp242: The argument `match` doesn't do anything in the function. We can remove it?

ktp242: Sorry I was wrong after reading more into the document. Please ignore the comment.

EdwinGuzman: It's a defined param for the callback function, but it is weird that it's not being used in this function itself. Removing it may break the functional call, even if it's not being used.

EdwinGuzman: @gkallenberg this could go in the `utils`.

EdwinGuzman:
I keep being redirected to the 404 page when going to `/best-books/staff-picks`:

`Status Code: undefined, Error Message: undefined` and `Type '404' not found in selectData`

PR #208, opened 2018-05-16
Updating the route to use the config's base url.

EdwinGuzman:
For RP


EdwinGuzman:
Hmmm I _think_ it's okay now, but it'll be good to have the fallback just in case. Since you've been communicating with Brett, I think you can decide and let him know what you think.

PR #212, opened 2018-05-17
Updating h2 to h3 in the side for Filter by Tags heading, and updatin…

EdwinGuzman:
…g main h2 heading so that the component only gets rendered when appropriate in mobile/desktop.

Jira: https://jira.nypl.org/browse/WWW-264
Up on dev: http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-movies/recommendations/best-books/staff-picks

- "Filter by Tags" is now an h3.

Things I want to test:
- Added an `aria-labelledby` attribute to each `li`. I do not want to remove the `tabIndex="0"` because we are using that to focus on each book card when using the anchored hash approach. Is this approach okay?
- Not rendering both H2s with ids `sidebar-list-title` and `list-title`. Instead, when it's in "mobile" only the `sidebar-list-title` will render and when in desktop the `list-title` will render. This, I believe, fixes half the issue. The other half is that in desktop, `list-title` is outside of the DOM heading structure. So the h3s in the sidebar will be tabbed through first before the H2. It is a difficult engineering task to get that same working element on mobile to display nicely in desktop. That is why we made a different one for desktop.
-- I did try to switch the DOM order for the sidebar and booklist and use flex to order them visually. It works but you will then have to tab through every book card before you can even reach the sidebar filters. This doesn't seem ideal to me. So re-grouping on this would be great.

For now, please check the code and let me know what you think about the approach of checking the document's width to render/not render the H2 heading components.


EdwinGuzman:
Update: Made it so there's only one `ListTitle` component above the `Sidebar` and `BookList` components and just added lots of CSS to manipulate it 😭 😭 😭 

It's working but I don't like it. We need to heavily cross-browser test it.

EdwinGuzman:
`aria-labelledby` is the correct approach according to https://www.w3.org/TR/wai-aria-1.1/#aria-label since I'm assuming that the title of the book is what we want to use and that is already being displayed on the page. Let me know if this is right @ricardogalvez @wlla .

EdwinGuzman:
Updated PR with new DOM structure for the H2 and tabbing flow.

src/app/components/ListTitle/ListTitle.jsx
```diff
@@ -6,26 +6,24 @@ const ListTitle = (props) => {
...
 
-  if (props.displayType !== 'staff-picks') {
-    return (
-      <h2 id={`${idPrefix}list-title`} tabIndex="0">
-        {displayDate.year} Picks
-        <span className="pick-count">{booksfound}</span>
-      </h2>
-    );
-  }
+  const isStaffPicks = (props.displayType !== 'staff-picks');
```
ktp242: The name and what it represents is opposite. If `props.displayType` is `ya`, `isStaffPicks` will be true.

ktp242:
Looks good. Just a minor change I think is needed.

src/app/components/ListTitle/ListTitle.jsx
EdwinGuzman: Good catch.

EdwinGuzman:
Updated.

src/app/components/Book/Book.jsx
```diff
@@ -63,21 +65,19 @@ const Book = ({ pick, isJsEnabled }) => {
         href={catalogUrl}
         className="catalog-url"
         onClick={() => gaEvent('Book')}
-        aria-label={`Request Book: ${book.title}`}
       >
         <BookIcon width="32px" height="32px" ariaHidden />
-        <span>{config.requestUrlsText.catalog}</span>
+        <span aria-label={`Request Book: ${book.title}`}>{config.requestUrlsText.catalog}</span>
```
gkallenberg: Are we using `aria-label` or `aria-labelledby` here for the book titles? I see `aria-labelledby` used for tags, but not in the book items themselves. Is that correct?

EdwinGuzman:
@ricardogalvez @wlla merging this in.

src/app/components/Book/Book.jsx
EdwinGuzman: Yeah, using `aria-label` to be more specific for the two request buttons.

Using `aria-labelledby` for the book cards since I'm linking the h3 id to the book card.

PR #213, opened 2018-05-17
Kp/www 268 put tags in ul

ktp242:
This PR wraps the tags inside a `<ul>` and updates each of them into an `<li>`. Because of the change, there are some DOM and styles need to be updated too. Of course with related tests.

However, as we add a `<ul>` here, the wrapper of `<ul>` can't be a `<p>` anymore. And it affects the way the line of tags breaks when there's not enough width of the column.

With enough width
screen shot 2018-05-17 at 3 28 36 pm

Without enough width
screen shot 2018-05-17 at 3 30 52 pm


courtneymcgee:
@ktp242 - Can we add &nbsp; between tags? Does that help? Apologies for the naïve question!

ktp242:
@courtneymcgee what kind of visual do you want to achieve? A bigger white space between `Drama,` and `On the dark side,` and etc? If so, we can totally do it either by adding more white spaces or just margin to style rules.

src/app/components/Book/Book.jsx
```diff
@@ -114,11 +114,12 @@ const Book = ({ pick, isJsEnabled }) => {
 
   const renderTags = (tags, jsEnabled) => {
     const tagsMarkup = !_isEmpty(tags) ?
-      tags.map((tag, i) => <span key={i}>{tag}{i !== (tags.length - 1) ? ', ' : ''}</span>) : null;
+      tags.map((tag, i) => <li key={i}><span>{tag}{i !== (tags.length - 1) ? ', ' : ''}</span></li>) : null;
```
EdwinGuzman: You don't need the `span`. You can add the CSS for the span directly to the li.

test/unit/Book.test.js
```diff
@@ -160,10 +169,29 @@ describe('Book Component', () => {
...
+      const tagsWrapper = component.find('.book-item-tags').find('ul');
+
+      expect(tagsWrapper.length).to.equal(1);
+      expect(tagsWrapper.find('li').length).to.equal(2);
+      expect(tagsWrapper.find('li').at(0).text()).to.equal('Offbeat, ');
+      expect(tagsWrapper.find('li').at(1).text()).to.equal('Seriously good writing');
+    });
+
+    it('should render the pick tags and its class should not be visuallyHidden.',
+      () => {
```
EdwinGuzman: This shouldn't be on a new line since it fits in the live above.

EdwinGuzman:
I'd say the updated margin which is there is more maintainable than the `&nsbp;`.

ricardoom:
definitely better to add margin than a `&nbsp;`

ricardoom:
oops

src/app/components/Book/Book.jsx
gkallenberg: What exactly is the function of the `span` tags here and around the "Tags" heading?

ktp242:
Should we just have the tags on a separate line then? @ricardoom

PR #217, opened 2018-05-18
Adding endpoints accessed to error log message for troubleshooting.

holingpoon:
Sample log message:
```
{"timestamp":"2018-05-18T20:52:07.825Z","levelCode":3,"level":"ERROR","pid":"11066","message":"Status Code: 404, Error Message: staff-picks/2019-03 not found, Endpoint: /book-lists?type=staff-picks, Redirecting to: /books-music-movies/recommendations/best-books/404"}
```


src/server/routes/annualData.js
```diff
@@ -123,7 +123,7 @@ function annualListData(req, res, next) {
 
       // If error returned from the endpoint
       if (data.statusCode >= 400) {
-        logger.error(`Status Code: ${data.statusCode}, Error Message: ${data.error}`);
+        logger.error(`Status Code: ${data.statusCode}, Error Message: ${data.error}, Endpoint: ${platformConfig.endpoints.annualPath}${dataType}/${requestedYear}`);
```
EdwinGuzman: You're missing two `.error` logs in this file.

src/server/routes/monthData.js
```diff
@@ -195,7 +195,7 @@ function selectClientMonthData(req, res) {
       });
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Endpoint: ${platformConfig.endpoints.staffPicksPath}${seasonMatches[0]}`);
```
EdwinGuzman: You're missing a `.error` log call in this file.

src/server/routes/annualData.js
```diff
@@ -69,7 +69,7 @@ function annualCurrentListData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Endpoint: ${platformConfig.endpoints.annualLists[dataType]},Redirecting to: ${config.baseUrl}/404`);
```
ktp242: This `catch` actually will catch the errors both from the first call for the lists and the second call for the books. I am not sure how to do it, but it could be incorrect if we always return the same endpoint for all the errors.

Same as the other one in `src/server/routes/annualData.js` and the other two in `src/server/routes/monthData.js`

holingpoon: Good point. It would make more sense to say which function was generating this error?

src/server/routes/annualData.js
```diff
@@ -69,7 +69,7 @@ function annualCurrentListData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Source: annualData.annualCurrentListData, Redirecting to: ${config.baseUrl}/404`, error);
```
ktp242: We have `source` here that does helps us locate the function, but we still don't know which endpoint caused the error. The `error` here is not logged out either.

I tired console `error`, and it should have the info as below

```
{ statusCode: 400,
  data: '{"error":"invalid_client","error_description":"The client credentials are invalid"}' }
```

While some useful info we might want to log, unfortunately it won't give us any endpoint info back.

holingpoon: I gave the app the wrong credentials to log into Platform API, but I get the following error in log
```
{"timestamp":"2018-05-18T20:57:48.571Z","levelCode":3,"level":"ERROR","pid":"11644","message":"Status Code: 400, Error Message: undefined, Endpoint: /book-lists?type=teens, Redirecting to: /books-music-movies/recommendations/best-books/404"}
```
It says Error Message is undefined. Any pointers?

ktp242: because there is no `error.code` anymore. If you tried to console `error` as I did, the error description is actually nested in `data`.

ktp242:
Probably a larger issue this one.
From what I found, if I used a incorrect endpoint at either of the calls that `annualData.annualCurrentListData` makes. It would not go int to `catch`, instead, it considered they were successful calls, and thus returned no error, instead of `response` with a message told me I did not provide correct OAuth secrets. 

And then it will render either empty list for the dropdown menu or empty list of books.

screencapture-local-nypl-org-3001-books-music-movies-recommendations-best-books-ya-2018-05-18-17_12_28

screencapture-local-nypl-org-3001-books-music-movies-recommendations-best-books-ya-2018-05-18-17_17_02

src/server/routes/annualData.js
```diff
@@ -69,7 +69,7 @@ function annualCurrentListData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Source: annualData.annualCurrentListData, Redirecting to: ${config.baseUrl}/404`, data);
```
EdwinGuzman: This is't right because `data` is defined in the previous `.then()` clause but not in this `.catch()` clause.

src/server/routes/annualData.js
```diff
@@ -163,11 +162,11 @@ function annualClientListData(req, res) {
   const yearMatches = matchListDate(req.params.time, req.params.type);
 
   if (!yearMatches) {
-    logger.error('Status Code: 400, Error Message: Invalid year.');
+    logger.error(`Status Code: 400, Error Message: Invalid year ${req.params.time} for type ${dataType}`);
 
     res.json({
       statusCode: 400,
-      errorMessage: 'Invalid year.',
+      errorMessage: `Invalid year ${req.params.time} and time ${req.params.type}`,
```
EdwinGuzman: Do you mean `type` instead of `time`?

src/server/routes/annualData.js
```diff
@@ -146,7 +145,7 @@ function annualListData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Endpoint: ${platformConfig.endpoints.annualLists[dataType]}, Redirecting to: ${config.baseUrl}/404`, error);
```
ktp242: I think this should also be `source` instead of `endpoint`? Following the same pattern of `annualCurrentListData`?

src/server/routes/monthData.js
```diff
@@ -62,7 +62,7 @@ function currentMonthData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Redirect to: ${config.baseUrl}/404`, error);
```
ktp242: This should have `source` too?

src/server/routes/monthData.js
```diff
@@ -157,7 +157,7 @@ function selectMonthData(req, res, next) {
       next();
     })
     .catch((error) => {
-      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}`);
+      logger.error(`Status Code: ${error.statusCode}, Error Message: ${error.code}, Redirecting to: ${config.baseUrl}/404`, error);
```
ktp242: Should this have `source` as well?

PR #218, opened 2018-05-18
Updating regex date pattern checker.

EdwinGuzman:
Just in case the length of the date in the returned data is more than 7 characters in length, then update the regex. A check for possible bad user input in the spreadsheet.


ktp242:
Do we have a test for this function? `matchListDate(dateStr, type)`?

EdwinGuzman:
Updated, added tests. Quick review?

PR #219, opened 2018-05-18
Small css update to fix non issue.

EdwinGuzman:
blah


ktp242:
I am wondering if we find or have even longer description in the future, we will have to update it again.

EdwinGuzman:
Same. I said future ones will be a "won't fix". This is where Ricardo's updated designs will fix the issue.

PR #227, opened 2018-05-22
Fixes #225; fix invalid variable values for log messages

### LINKED ISSUE #225: Invalid Direct Title Links Lead to an Internal Server Error
Direct staff-picks links lead to an "Internal Server Error" when an invalid, non-date path is added after /books-music-movies/recommendations/best-books/staff-picks such as: /books-music-movies/recommendations/best-books/staff-picks/foo

gkallenberg:
Fixes an issue with log message variables being passed incorrectly for monthData.selectMonthData. Season matches were being passed down to the log which, when invalid, caused an error instead of redirecting to a 404. Audience was also updated here to pass the requested value.

### DIFF
--- a/src/server/routes/monthData.js
+++ b/src/server/routes/monthData.js
@@ -99,7 +99,7 @@ function selectMonthData(req, res, next) {
   }
 
   if (!seasonMatches || !isValidAudience) {
-    let matches = 'undefined';
+    let matches;
     if (Array.isArray(seasonMatches) && seasonMatches.length >= 1) {
       matches = seasonMatches[0];
     }
--- a/src/server/routes/monthData.js
+++ b/src/server/routes/monthData.js
@@ -68,9 +68,10 @@ function currentMonthData(req, res, next) {
     });
 }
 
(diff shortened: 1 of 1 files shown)


src/server/routes/monthData.js
```diff
@@ -86,24 +87,23 @@ function selectMonthData(req, res, next) {
...
-    } else {
-      // Or set it is an invalid audience query
-      isValidAudience = false;
-    }
+    isValidAudience = ['Adult', 'YA', 'Children'].includes(audience);
   }
 
   if (!seasonMatches || !isValidAudience) {
-    logger.error(`Status Code: 400, Error Message: Invalid season ${seasonMatches[0]} or audience ${audience}, Redirecting to: ${config.baseUrl}/404`);
+    let matches = 'undefined';
```
EdwinGuzman: Why do you have to make this a string?

gkallenberg: Was trying to assign as undefined, but I can probably just assign no value instead which will result in undefined.

PR #230, opened 2018-06-01
404, about, and title changes

gkallenberg:
Corresponds to the adjustments requested as part of https://jira.nypl.org/browse/WWW-317

* Adds line with link to Staff Picks on the 404 page template
* Includes more text in the About page link to print formats
* Updates titles to use `Adults` and `Young Adults` plural forms
** was necessary to add some CSS to adjust for longer titles so the # found would not wrap
** see: http://localhost:3001/books-music-movies/recommendations/best-books/staff-picks/2015-09?audience=YA
** reset _top_heading.scss `h2#list-title` to use the default `font-size: 1.6em` and remove `letter-spacing: -0.3px` in order to trigger the wrapping
** wrapping occurs on lists for Young Adults with Summer, September, October, November in the title.
** CSS change fixes for the longest case of September

### DIFF
--- a/src/app/components/Error404Page/Error404Page.jsx
+++ b/src/app/components/Error404Page/Error404Page.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import config from '../../../../appConfig';
 
 const Error404Page = () => (
   <div className="error-container">
@@ -8,7 +9,7 @@ const Error404Page = () => (
     </p>
     <p>
       Ready to discover your next great read? Browse and filter our&nbsp;
-      <a href="/books-music-movies/recommendations/best-books/staff-picks">
+      <a aria-label="Browse and filter our Staff Picks" href={`${config.baseUrl}/staff-picks`}>
         Staff Picks.
       </a>
     </p>
--- a/src/app/components/Error404Page/Error404Page.jsx
+++ b/src/app/components/Error404Page/Error404Page.jsx
@@ -15,7 +15,7 @@ const Error404Page = () => (
     </p>
     <p>
       Need help or have a question?&nbsp;
-      <a href="/get-help/contact-us">
+      <a href="https://www.nypl.org/get-help/contact-us">
         Contact us.
       </a>
     </p>
--- a/src/app/components/Error404Page/Error404Page.jsx
+++ b/src/app/components/Error404Page/Error404Page.jsx
@@ -7,8 +7,14 @@ const Error404Page = () => (
       We’re sorry. The page you were looking for doesn’t exist.
     </p>
     <p>
+      Ready to discover your next great read? Browse and filter our&nbsp;
+      <a href="/books-music-movies/recommendations/best-books/staff-picks">
+        Staff Picks.
+      </a>
+    </p>
+    <p>
       Need help or have a question?&nbsp;
-      <a href="https://www.nypl.org/get-help/contact-us">
+      <a href="/get-help/contact-us">
         Contact us.
       </a>
     </p>
--- a/test/unit/Error404Page.test.js
+++ b/test/unit/Error404Page.test.js
@@ -23,24 +23,35 @@ describe('Error404Page', () => {
       expect(title.text()).to.equal('404 | Not Found');
     });
 
-    it('should have two <p>s with correct contents.', () => {
+    it('should have three <p>s with correct contents.', () => {
       const descriptionParagraphs = component.find('p');
 
-      expect(descriptionParagraphs.length).to.equal(2);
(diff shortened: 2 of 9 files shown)


test/unit/utils.test.js
```diff
@@ -222,4 +222,15 @@ describe('Utils functions', () => {
       }
     );
   });
+
+  describe('makeAgeGroupPlural', () => {
+    it('should add an `s` to the end of the given word', () => {
+      expect(utils.makeAgeGroupPlural('Adult')).to.eql('Adults');
+      expect(utils.makeAgeGroupPlural('Young Adult')).to.eql('Young Adults');
+      expect(utils.makeAgeGroupPlural('cat')).to.eql('cats');
```
ktp242: nice!

ktp242:
Approved! However, I have a question which is not related to dev's work. I have addressed it on the ticket https://jira.nypl.org/browse/WWW-300

After this is merged, we should update the CHANGELOG and version as well.

src/app/components/Error404Page/Error404Page.jsx
```diff
@@ -6,9 +6,15 @@ const Error404Page = () => (
...
+    <p>
+      Ready to discover your next great read? Browse and filter our&nbsp;
+      <a href="/books-music-movies/recommendations/best-books/staff-picks">
+        Staff Picks.
+      </a>
+    </p>
     <p>
       Need help or have a question?&nbsp;
-      <a href="https://www.nypl.org/get-help/contact-us">
+      <a href="/get-help/contact-us">
```
EdwinGuzman: Just fyi, this won't work in the `staff-picks-development.....com` domains but will work fine in the reverse proxy app.

test/unit/Error404Page.test.js
```diff
@@ -23,24 +23,35 @@ describe('Error404Page', () => {
...
+      expect(descriptionParagraphs.at(2).text()).to.equal(
         'Need help or have a question? Contact us.'
       );
     });
 
+    it('should have a link to "Staff Picks".', () => {
+      const askNyplLink = component.find('a');
+
+      expect(askNyplLink.length).to.equal(2);
+      expect(askNyplLink.nodes[0].props.href).to.equal('/books-music-movies/recommendations/best-books/staff-picks');
```
EdwinGuzman: The baseUrl config variable should be used instead of the string.

test/unit/utils.test.js
EdwinGuzman: What if you do `utils.makeAgeGroupPlural('goose')`? :) jk it's fine.

test/unit/utils.test.js
```diff
@@ -222,4 +222,15 @@ describe('Utils functions', () => {
       }
     );
   });
+
+  describe('makeAgeGroupPlural', () => {
```
EdwinGuzman: What happens if you pass in a number to `utils.makeAgeGroupPlural`? Or don't pass anything at all? For the latter, I think you'd get `undefined` as the returned value.

EdwinGuzman:
I merged the footer change into `development`. Can you just include your updates in the changelog, please?

ktp242:
I tried to do a fresh `npm install` for this, but I got an error from running it after install finished successfully
```
common.js:12 Uncaught Error: Cannot find module "fs"
    at webpackMissingModule (common.js:12)
    at Object.eval (common.js:12)
    at eval (common.js:497)
    at Object.<anonymous> (bundle.js:1447)
    at __webpack_require__ (bundle.js:670)
    at fn (bundle.js:96)
    at eval (winston.js:24)
    at Object.<anonymous> (bundle.js:3030)
    at __webpack_require__ (bundle.js:670)
    at fn (bundle.js:96)
```

src/app/components/Error404Page/Error404Page.jsx
gkallenberg: My attempt to avoid adding absolute links to apps. If this is not wise, let me know if I should revert.

gkallenberg:
Fix the issue with compilation errors, the project should build correctly again. As a side note, there are CSS changes which affect letter kerning and may need review from @ricardogalvez. The ...Staff Picks for Young Adults... titles have been tweaked to reduce letter spacing for longer titles that include `Young Adults` and longer month names like `September` and `November`

src/app/components/Error404Page/Error404Page.jsx
EdwinGuzman: This is fine but we'll probably get some "it's not working"  feedback if anyone tests the non reverse proxy urls.

PR #231, opened 2018-06-06
Small CSS override to the footer list item.

EdwinGuzman:
Normalize screws up the footer social media icons only on this app so I made this PR in the repo instead of the footer package.

dev: http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-movies/recommendations/best-books/staff-picks

### DIFF
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -50,3 +50,8 @@ body {
 li {
   list-style: none;
 }
+
+// Small override to a normalize rule
+.footer-content .socialMedia li {
+  margin-left: 5;
+}
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -69,4 +69,10 @@ li {
   .copyright p {
     max-width: 100%;
   }
+
+  &-content {
+    @include min-screen(768px) {
+      left: auto;
+    }
+  }
 }
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -51,7 +51,8 @@ li {
   list-style: none;
 }
 
-// Small override to a normalize rule
+// Small override to a normalize rule which mess up the layout for
+// the footer in the mobile and tablet viewports.
 .footer {
   .socialMedia li {
     margin-left: 0;
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -52,6 +52,20 @@ li {
 }
 
 // Small override to a normalize rule
-.footer-content .socialMedia li {
-  margin-left: 0;
+.footer {
+  .socialMedia li {
+    margin-left: 0;
+  }
+
+  .footerLinks {
+    width: 81%;
+
+    @include min-screen(768px) {
+      width: 100%;
+    }
+  }
+
+  .copyright p {
+    max-width: 100%;
+  }
 }
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -53,5 +53,5 @@ li {
 
 // Small override to a normalize rule
 .footer-content .socialMedia li {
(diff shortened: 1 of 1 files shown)


ktp242:
The code look good. Well done. Any place can I go to see the differences? And do we know why it only messes up in Staff Picks?

EdwinGuzman:
You can see it in the link in the description...
Yes, I also wrote in the description that it's because of the normalize stylesheet that messes things up.

ktp242:
@EdwinGuzman Thanks for pointing out!
@ricardoom Thanks. I think you are right!

Is that possible we have a fix on DTL that can solve the problem instead of a single fix on Staff Picks? This way if we update Footer in the future, we don't have to particularly check Staff Picks again.

The other thing, I found on dev http://staff-picks-development.us-east-1.elasticbeanstalk.com/books-music-movies/recommendations/best-books/staff-picks
at the viewport 614-766px wide, the links below social media icons are not at the right position as well.

ricardoom:
@ktp242 please provide screen grabs and browser info.

ktp242:
Browser is chrome version 66.
If you go to the link from the comment, open the dev tool, and shrink the width of the page to 753px you should be able to see it.

ricardoom:
@ktp242 thanks for the tips on how to use a browser inspector. I'm curious as to what _you're_ seeing and consider to be an error. While there is some layout reflow issues, I am trying to figure out if it warrants the amount of work, testing and deploying for a 152px range...

ricardoom:
While we're at it, the 1024 ~ 1040, has a more pressing issue

screen shot 2018-06-06 at 4 38 52 pm
screen shot 2018-06-06 at 4 38 38 pm

ktp242:
Got it! I can definitely help you get a better idea.

screen shot 2018-06-06 at 4 44 01 pm

And seeing what you and I have found, I feel we should investigate why Staff Picks behaves so different than other apps and solve the problem(s) there. So in the future, if we make any changes on Footer, we don't have to worry about Staff Picks acting differently again.

EdwinGuzman:
Yes, that's fine just please read the description to the PR.

The investigation is done and is what we've said about the design toolkit messing it up with the rule that Ricardo gave. Yes, I do agree that the footer should be stand alone and if we can make changes to the footer and not the app that would be great. But again, this is the nypl normalize stylesheet which covers the global `li` rule so there's no going around it except in the app. The better approach will be to update that normalize css stylesheet.

I'll update this PR to fix that mobile issue.

ktp242:
Cool. I didn't know we have investigated this.

If this is something that would cause bigger issue and no other way around, then I guess fix on Staff Picks is what has to be done.

Since we won't update normalize css stylesheet any time soon, let's have a better documentation somewhere in the app.

And thanks for going to fix the mobile issue!

EdwinGuzman:
Updated.

ktp242:
Thanks for the update! I noticed the right edge align of the social media icons is a bit off with the copyright texts. See the image below,

screen shot 2018-06-07 at 12 59 28 pm

And here is the one from Discovery
screen shot 2018-06-07 at 1 00 56 pm

Not sure if it is worth to be fixed.

And I still see the issue Ricardo mentions at the viewport of 1024 ~ 1040. Again, I am not sure if it needs to be fixed in this PR. 

I think both of the fixes will be @ricardoom 's call.

EdwinGuzman:
Great, fixed the issue @ricardogalvez  brought up.
With rem and percentages, it won't line up perfectly. That's good enough and not worth aligning it in my opinion. Again, these are overrides that only affect the Footer in Staff Picks and we should eventually fix the normalize stylesheet to get what we actually want.

PR #232, opened 2018-06-07
Adjustments for 404 and About components

gkallenberg:
* Reverted a change to the scss to avoid wrapping the "Books Found" portion on longer titles as per Ricardo's recommendation
* Removed the aria-label added to the 404 page line linking back to the Staff Picks main page
* Updated the text for additional information in the About.jsx component as per copy changes in [WWW-299](https://jira.nypl.org/browse/WWW-299)

### DIFF
--- a/src/app/components/About/About.jsx
+++ b/src/app/components/About/About.jsx
@@ -43,8 +43,10 @@ class About extends React.Component {
     return (
       <div className="about nypl-row">
         <h2>Additional Information</h2>
-        <p>Many of these titles are available in&nbsp;
-          <a href={aboutUrls.print}>formats for patrons with print disabilities.</a>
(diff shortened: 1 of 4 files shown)


ktp242:
Looks good. So it is OK to have two lines for the title? Like the one from `staff-picks/2015-09`, `young adults`?

gkallenberg:
Wrapping lines is fine according to Ricardo. The containing h2 CSS is provided by normalize and doesn't expand to the containing div's width. The best approach would probably be to fix that at some point, but it's not part of this PR's mandate. The wrapping can be addressed in another issue/PR elsewhere, I think.

ktp242:
Cool thanks for that!

PR #234, opened 2018-06-25
Updated header to 2.4.9

ktp242:
This PR just updates the header to 2.4.9. Can be reviewed on dev
https://dev-www.nypl.org/books-music-movies/recommendations/best-books/staff-picks

The main change should be the visible focus outline styles. It should be dark blue color with shadows
as the screenshot shows
screen shot 2018-06-25 at 2 29 36 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## CHANGELOG
 
+### 4.0.2
+> Update the header component to 2.4.9.
+
 ### 4.0.1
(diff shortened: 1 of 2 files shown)


gkallenberg:
The header styles seem to be good for the focus changes to the dark blue, shadowed boxes. Did we lose focus styles around the cards? I only see dark blue focus rings on the reserve buttons. No visible focus is shown around the cards themselves which I thought we added at one point.

ktp242:
hmmm I am not sure. I don't recall we have imported the latest design toolkit to Staff-Picks. Maybe that is the reason?

